PR #2, opened 2023-07-24
Create ECR in Meta

Chirag-Bhatti:


terraform/modules/ecr/main.tf
```diff
@@ -0,0 +1,40 @@
...
+      action = {
+        type = "expire"
+      }
+      selection = {
+        tagStatus   = "any"
+        countType   = "imageCountMoreThan"
+        countNumber = 10
+      }
+    }]
+  })
```
RachaelBooth, line 39: This needs thinking through as well - I expect to need to keep more than the last 10 images, since we're likely to build more than that between releases to production (particularly with review apps). 

I expect this is something to have a final answer for when thinking about tagging strategy and deployments (the ideal would probably be something clever that treats different envs differently), but for now I think it would be better to have nothing (because the consequences of not then changing it are better that way).

terraform/modules/ecr/main.tf
```diff
@@ -0,0 +1,40 @@
...
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~>5.0"
+    }
+  }
+}
+
+#tfsec:ignore:aws-ecr-repository-customer-key:encryption using KMS CMK not required
+resource "aws_ecr_repository" "main" {
```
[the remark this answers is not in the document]
cschiragb: Updated the name to `_` 👍

terraform/modules/ecr/main.tf
```diff
@@ -0,0 +1,40 @@
...
+  #checkov:skip=CKV_AWS_136:encryption using KMS not required
+  name                 = "core-ecr"
+  image_tag_mutability = "IMMUTABLE"
```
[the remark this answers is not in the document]
cschiragb: I've updated this to `MUTABLE` so we expect this to be initially most useful 😀

terraform/modules/ecr/main.tf
cschiragb: Ah yes that makes sense. I've removed the policy all together for now and can add one in later once we have a clearer idea. Happy to make a ticket so we don't forget to make one later?

terraform/modules/ecr/main.tf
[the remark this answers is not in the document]
cschiragb: Unfortunately because of this change the checkov and tfsec tools complain as they prefer `IMMUTABLE`, so I had to add ignore statements too 😅

terraform/modules/ecr/main.tf
RachaelBooth: There already is one in the deployment pipelines epic.

terraform/modules/ecr/main.tf
```diff
@@ -0,0 +1,23 @@
...
+      version = "~>5.0"
+    }
+  }
+}
+
+#tfsec:ignore:aws-ecr-enforce-immutable-repository:we require mutable tags for our current deployment process
+#tfsec:ignore:aws-ecr-repository-customer-key:encryption using KMS CMK not required
+resource "aws_ecr_repository" "this" {
+  #checkov:skip=CKV_AWS_51:we require mutable tags for our current deployment process
+  #checkov:skip=CKV_AWS_136:encryption using KMS not required
```
RachaelBooth, line 16: I assume there's no way around needing to repeat the skip / ignore comments in slightly different places? It would be nicer if we could group ones covering the same thing, but if they need to be in different spots then so be it. (It is a major annoyance about using both though)

terraform/modules/backend/main.tf
```diff
@@ -198,7 +198,7 @@ module "tf_state_replica_log_bucket" {
 #tfsec:ignore:aws-s3-no-public-buckets:restrict_public_buckets is set to true, this is a false flag
 #tfsec:ignore:aws-s3-encryption-customer-key:using key managed by S3 because cloudposse backend module can't be configured with KMS
 #tfsec:ignore:aws-s3-enable-versioning:versioning not required for replica of a bucket with versioning enabled
-#tfsec:ignore:aws-s3-enable-bucket-logging:exp:2023-07-24:ignoring this warning temporarily while deciding if logging is necessary
+#tfsec:ignore:aws-s3-enable-bucket-logging:access logs are being taking, this is a false flag
```
RachaelBooth, line 201: `access logs are being taken, ...` I think?

RachaelBooth (approved):
Minor thing about comment wording, but I don't think this needs to come back to me.

terraform/modules/ecr/main.tf
cschiragb: I believe the only way around avoiding this sort of repetition is by maybe providing the ignore/skip statements at a higher level where appropriate (e.g. where the module is being called from)

I think in general we will aim to put all ignore/skip statements on the module declaration itself, so that anything that calls it won't need to provide such statements

I think this is more of an issue with the way the code is structured for creating the backend, but the above approach I think will serve fine going forward for the rest of the infra

We could create a wrapper for the 3 S3 buckets declared in backend/main.tf, and put the statements on the wrapper, but I don't think it's worth all the work. Are you happy for us to leave the backend as is?

RachaelBooth (approved):

PR #3, opened 2023-08-02
Set up networking

Chirag-Bhatti:
Implemented the following
- VPC with flow logs
- Subnets
- Internet and NAT gateways
- Routing

Also added any necessary ignore statements for checks made by tfsec and checkov


terraform/modules/networking/vpc.tf
```diff
@@ -0,0 +1,87 @@
...
+  iam_role_arn    = aws_iam_role.vpc_flow_logs.arn
+  log_destination = aws_cloudwatch_log_group.vpc_flow_logs_rejected.arn
+  traffic_type    = "REJECT"
+  vpc_id          = aws_vpc.main.id
+}
+
+# tfsec:ignore:aws-cloudwatch-log-group-customer-key:flow logs are non-sensitive
+resource "aws_cloudwatch_log_group" "vpc_flow_logs_accepted" {
+  #checkov:skip=CKV_AWS_158:flow logs are non-sensitive
+  #checkov:skip=CKV_AWS_338:we think that log retention of at least 1 year is excessive for this application and are using 90 days
```
RachaelBooth, line 29: Might be worth rewording this comment since we don't define the length in place here, but I'm not that fussed.

RachaelBooth (approved):
I think this has gone a bit beyond the ACs here (specifically the logging stuff, which is wanted at some point but maybe not as minimum viable networking). No need to change anything in this work, but it may be worth having a chat for future tickets (either as an HLD-type stage or at refinement) to make sure we're aligned on what's needed.

hugh-emerson:
LGTM 👍

hugh-emerson (approved):

PR #4, opened 2023-08-14
CLDC-2527: Setup RDS

Chirag-Bhatti:
- Implement the RDS terraform module (excluding any restorability / backups implementation)
- Configure RDS module for all envs

### DIFF
--- a/terraform/development/main.tf
+++ b/terraform/development/main.tf
@@ -34,13 +34,13 @@ module "networking" {
   vpc_flow_cloudwatch_log_expiration_days = 90
 }
 
-module "rds" {
+module "database" {
   source = "../modules/rds"
 
-  prefix               = "core-dev"
-  allocated_storage    = 5
-  db_subnet_group_name = module.networking.private_subnet_group_name
-  instance_class       = "db.t3.micro"
-  security_group_ids   = []
-  vpc_id               = module.networking.vpc_id
+  prefix                            = "core-dev"
+  allocated_storage                 = 5
+  db_subnet_group_name              = module.networking.private_subnet_group_name
+  instance_class                    = "db.t3.micro"
+  ingress_source_security_group_ids = []
+  vpc_id                            = module.networking.vpc_id
 }
--- a/terraform/modules/rds/security_groups.tf
+++ b/terraform/modules/rds/security_groups.tf
@@ -5,7 +5,7 @@ resource "aws_security_group" "rds" {
 
   ingress {
     description     = "Allow ingress on port 5432 only from the specified security groups"
-    security_groups = var.security_group_ids
+    security_groups = var.ingress_source_security_group_ids
     from_port       = 5432
     to_port         = 5432
     protocol        = "tcp"
--- a/terraform/modules/rds/variables.tf
+++ b/terraform/modules/rds/variables.tf
@@ -18,7 +18,7 @@ variable "prefix" {
   description = "The prefix to be prepended to resource names."
 }
 
-variable "security_group_ids" {
+variable "ingress_source_security_group_ids" {
   type        = list(string)
   description = "The security group ids (sources) the rds security group will allow ingress from"
 }
--- a/terraform/production/main.tf
+++ b/terraform/production/main.tf
@@ -34,13 +34,13 @@ module "networking" {
   vpc_flow_cloudwatch_log_expiration_days = 90
 }
 
-module "rds" {
+module "database" {
   source = "../modules/rds"
 
-  prefix               = "core-prod"
-  allocated_storage    = 100
-  db_subnet_group_name = module.networking.private_subnet_group_name
-  instance_class       = "db.t3.small"
-  security_group_ids   = []
-  vpc_id               = module.networking.vpc_id
+  prefix                            = "core-prod"
+  allocated_storage                 = 100
+  db_subnet_group_name              = module.networking.private_subnet_group_name
+  instance_class                    = "db.t3.small"
+  ingress_source_security_group_ids = []
+  vpc_id                            = module.networking.vpc_id
 }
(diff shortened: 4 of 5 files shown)


terraform/modules/rds/database.tf
```diff
@@ -0,0 +1,34 @@
...
+  #checkov:skip=CKV2_AWS_60:copy tags to snapshots TODO CLDC-2261
+  identifier                 = "${var.prefix}-postgres-db"
+  apply_immediately          = true
+  auto_minor_version_upgrade = true
+  allocated_storage          = var.allocated_storage #units are GiB
+  db_name                    = "data_collector"
+  db_subnet_group_name       = var.db_subnet_group_name
+  deletion_protection        = true # needs to be set to false and applied if you need to delete the DB
+  engine                     = "postgres"
+  engine_version             = "13.11"
```
Chirag-Bhatti, line 20: I think we're using Postgres 13.5 (looking at the existing pipelines etc). Closest version we can get is 13.7 on AWS, but I've taken 13.11 here as I assume the latest minor version will be the best while compatible

terraform/modules/rds/database.tf
```diff
@@ -0,0 +1,34 @@
+#tfsec:ignore:aws-rds-specify-backup-retention:TODO CLDC-2261
+#tfsec:ignore:aws-rds-enable-performance-insights:TODO CLDC-2260 if necessary
+#tfsec:ignore:AVD-AWS-0176:iam authentication not suitable as tokens only last 15minutes, password authentication preferred
```
Chirag-Bhatti, line 3: There seems to be a slight bug with `tfsec` here in that didn't output out a "worded id" of the rule it was flagging, so we've had to use a different reference (`AVD-AWS-0176`) to ignore it. [Here's a bug report](https://github.com/aquasecurity/tfsec/issues/1935) of a similar issue. One option could be to downgrade tfsec from `1.28.1` to `1.28.0` if we see many more like this in future.

terraform/modules/rds/security_groups.tf
```diff
@@ -0,0 +1,17 @@
+resource "aws_security_group" "rds" {
+  name        = "${var.prefix}-rds-security-group"
+  description = "RDS security group"
+  vpc_id      = var.vpc_id
+
+  ingress {
+    description     = "Allow ingress on port 5432 only from the specified security groups"
+    security_groups = var.security_group_ids
```
Chirag-Bhatti, line 8: We're not actually passing in `security_group_ids` here yet (this var is just passed in as `[]` for now) but will be in the near future e.g. once there's a security group for the ECS, we'll pass it in here so that ECS will be allowed "ingress" to the DB

terraform/staging/main.tf
```diff
@@ -33,3 +33,14 @@ module "networking" {
...
   vpc_flow_cloudwatch_log_expiration_days = 90
 }
+
+module "rds" {
+  source = "../modules/rds"
+
+  prefix               = "core-stag"
+  allocated_storage    = 5
+  db_subnet_group_name = module.networking.private_subnet_group_name
+  instance_class       = "db.t3.micro"
```
Chirag-Bhatti, line 43: PaaS uses `tiny-unecrypted-13` for staging and this looks like it is actually `db.t3.micro` in AWS so I've used this here (we check this using `cf marketplace -e postgres`)

terraform/development/main.tf
```diff
@@ -33,3 +33,14 @@ module "networking" {
...
   vpc_flow_cloudwatch_log_expiration_days = 90
 }
+
+module "rds" {
+  source = "../modules/rds"
+
+  prefix               = "core-dev"
+  allocated_storage    = 5
+  db_subnet_group_name = module.networking.private_subnet_group_name
+  instance_class       = "db.t3.micro"
```
Chirag-Bhatti, line 43: PaaS uses `tiny-unecrypted-13` for dev and this looks like it is actually `db.t3.micro` in AWS so I've used this here (we check this using `cf marketplace -e postgres`)

terraform/production/main.tf
```diff
@@ -33,3 +33,14 @@ module "networking" {
...
   vpc_flow_cloudwatch_log_expiration_days = 90
 }
+
+module "rds" {
+  source = "../modules/rds"
+
+  prefix               = "core-stag"
+  allocated_storage    = 100
+  db_subnet_group_name = module.networking.private_subnet_group_name
+  instance_class       = "db.t3.small"
```
Chirag-Bhatti, line 43: PaaS uses `small-ha-13` for prod and this looks like it is actually `db.t3.small` in AWS so I've used this here (we check this using `cf marketplace -e postgres`)

terraform/production/main.tf
```diff
@@ -33,3 +33,14 @@ module "networking" {
   vpc_cidr_block                          = "10.0.0.0/16"
   vpc_flow_cloudwatch_log_expiration_days = 90
 }
+
+module "rds" {
+  source = "../modules/rds"
+
+  prefix               = "core-prod"
+  allocated_storage    = 100
```
Chirag-Bhatti, line 41: Looks like prod had ~100GiB of free space a year ago on PaaS (not sure if this was at inception or well used, as that's the furthest I could look), so I assumed this is the fresh disk space we'll want in AWS at minimum

terraform/modules/rds/variables.tf
```diff
@@ -0,0 +1,29 @@
...
+  type        = string
+  description = "The instance class of the DB."
+}
+
+variable "prefix" {
+  type        = string
+  description = "The prefix to be prepended to resource names."
+}
+
+variable "security_group_ids" {
```
RachaelBooth, line 21: The description is clear enough, but if you can think of a way of encapsulating it better in this name it'd be worth it - as is it took looking at the description to think this was for allowing access to the db rather than what security group the db should be in.

RachaelBooth (approved):
Couple of comments on naming, otherwise looks fine for the moment (there's some things that'll need to be touched more in future tickets).

PR #5, opened 2023-08-14
CLDC-2533: Set up access to ECR for ECS task execution role

RachaelBooth:


terraform/modules/ecr/main.tf
```diff
@@ -21,3 +21,25 @@ resource "aws_ecr_repository" "this" {
...
   }
 }
+
+data "aws_iam_policy_document" "this" {
+  statement {
+    effect = "Allow"
+
+    principals {
+      type        = "AWS"
+      identifiers = formatlist("arn:aws:iam::%s:*", var.accessible_from_accounts)
```
RachaelBooth, line 31: Allows access for any user/role from these accounts. This may well be more than we want - restricting to the task execution role would cause some interesting headaches around ordering that it's not worth dealing with right now. (Apparently you can only apply one repository policy, so there's no nice getting around it by defining it with the role...)

RachaelBooth: Scratch that - AWS gives an annoyingly unhelpful error that goes away when you remove wildcards from this, so going with specifying the role arns.

terraform/meta/main.tf
```diff
@@ -43,4 +43,7 @@ module "prod_backend" {
 
 module "ecr" {
   source = "../modules/ecr"
+
+  # This will need updating to include dev and production roles
+  allow_access_by_roles = ["arn:aws:iam::107155005276:role/core-stag-task-execution"]
```
RachaelBooth, line 48: Hardcoding this is obviously annoying (was alright with the account ids, but this seems much more painful).

It'll do for now - will need to think through ordering and what dependencies we're happy to have where in order to do a nicer set up where these arns get picked up automatically, which we should do but can be at a later point.

RachaelBooth: https://dluhcdigital.atlassian.net/browse/CLDC-2675

terraform/modules/service/roles.tf
```diff
@@ -0,0 +1,24 @@
+# TODO: This may not be the best way to split files, consider in CLDC-2526
```
Chirag-Bhatti, line 1: I've added an AC to this ticket to check for TODOs, as we're not in the habit of checking for these yet 👍

terraform/modules/service/roles.tf
```diff
@@ -0,0 +1,24 @@
...
+      type        = "Service"
+      identifiers = ["ecs-tasks.amazonaws.com"]
+    }
+  }
+}
+
+resource "aws_iam_role_policy_attachment" "task_execution_managed_policy" {
+  role = aws_iam_role.task_execution.name
+  # This is an aws managed policy for ecs task execution roles
+  policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy"
```
Chirag-Bhatti, line 23: More for my benefit, I've generally felt it's good to avoid using AWS managed policies in case they change anything about them in future. However this one here seems pretty ideal for what we need 🤷‍♂️. What's your general preferences around these?

RachaelBooth: I think we generally do want to use AWS managed policies if there is one designed for our use case - precisely because if something changes about the permissions required we'll automatically get the update (e.g. if AWS adds a new permission needed to start ECS tasks). And there's less risk of getting something wrong.

This only applies if it is actually our use case rather than something tangentially related that happens to need similar permissions, but in this case it is.

terraform/meta/main.tf
```diff
@@ -43,4 +43,7 @@ module "prod_backend" {
 
 module "ecr" {
   source = "../modules/ecr"
+
+  # This will need updating to include dev and production roles
```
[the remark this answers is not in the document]
RachaelBooth: We'd need to create the roles in dev and prod, and I don't think we want to be applying anything there yet.

terraform/development/main.tf
```diff
@@ -44,3 +44,9 @@ module "database" {
   ingress_source_security_group_ids = []
   vpc_id                            = module.networking.vpc_id
 }
+
+module "service" {
+  source = "../modules/service"
+
+  prefix = "core-stag"
```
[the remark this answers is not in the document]
RachaelBooth: 🤦 - I almost didn't bother actually putting this in at all for dev because we'll need to re-evaluate it all come review apps anyway, obviously wasn't paying attention. (Maybe a separate chat to have about whether we should be bothering to fill in these files at the moment, but for now I'll update the prefix - and you're right it really should be a local variable)

terraform/meta/main.tf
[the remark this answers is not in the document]
Chirag-Bhatti: I think we might forget to add these in once we've created the roles in and dev and prod, however it will probably be quickly apparent what's missing so don't think it's worth creating a ticket / updating this comment to a TODO for

Chirag-Bhatti (approved):

Chirag-Bhatti:
I'll schedule a separate chat about creating modules for dev and prod, and whether we should be applying anything there yet 😀. Thanks for doing this ticket.

PR #6, opened 2023-08-15
CLDC-2528: Setup Redis

Chirag-Bhatti:
- Implement Redis without snapshots, logging and multi-az / failover
- Make some minor corrections and improvements to the codebase

### DIFF
--- a/terraform/modules/elasticache/parameter_groups.tf
+++ b/terraform/modules/elasticache/parameter_groups.tf
@@ -1,5 +1,5 @@
 resource "aws_elasticache_parameter_group" "main" {
-  name = "${var.prefix}-maxmemory-parameter-group"
+  name = "${var.prefix}-redis-parameter-group"
   # Ensure the redis version below tallies with the engine_version defined for the redis elasticache cluster (see redis.tf)
   # Parameter group families are outlined here - https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/ParameterGroups.Redis.html
   family = "redis6.x"
--- a/terraform/modules/elasticache/parameter_groups.tf
+++ b/terraform/modules/elasticache/parameter_groups.tf
@@ -1,0 +1,11 @@
+resource "aws_elasticache_parameter_group" "main" {
+  name = "${var.prefix}-maxmemory-parameter-group"
+  # Ensure the redis version below tallies with the engine_version defined for the redis elasticache cluster (see redis.tf)
+  # Parameter group families are outlined here - https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/ParameterGroups.Redis.html
+  family = "redis6.x"
+
+  parameter {
+    name  = "maxmemory-policy"
+    value = "volatile-lru"
+  }
+}
--- a/terraform/modules/elasticache/redis.tf
+++ b/terraform/modules/elasticache/redis.tf
@@ -1,6 +1,8 @@
-#tfsec:ignore:aws-elasticache-enable-backup-retention:TODO CLDC-XXXX setup a snapshot retention limit
+#tfsec:ignore:aws-elasticache-enable-backup-retention:TODO CLDC-2679 setup a snapshot retention limit
 resource "aws_elasticache_cluster" "main" {
-  #checkov:skip=CKV_AWS_134:TODO CLDC-XXXX setup a snapshot retention limit
+  #checkov:skip=CKV_AWS_134:TODO CLDC-2679 setup a snapshot retention limit
+  #TODO CLDC-2679 setup redis replicas with multi-az and automatic failover
+  #TODO CLDC-2682 setup redis logging
   cluster_id                 = "${var.prefix}-redis"
   auto_minor_version_upgrade = true
   apply_immediately          = true
--- a/terraform/modules/elasticache/redis.tf
+++ b/terraform/modules/elasticache/redis.tf
@@ -1,0 +1,16 @@
+#tfsec:ignore:aws-elasticache-enable-backup-retention:TODO CLDC-XXXX setup a snapshot retention limit
+resource "aws_elasticache_cluster" "main" {
+  #checkov:skip=CKV_AWS_134:TODO CLDC-XXXX setup a snapshot retention limit
+  cluster_id                 = "${var.prefix}-redis"
+  auto_minor_version_upgrade = true
+  apply_immediately          = true
+  engine                     = "redis"
+  engine_version             = "6.2"
+  maintenance_window         = "sun:23:00-mon:01:30"
+  node_type                  = var.node_type
+  num_cache_nodes            = 1
+  parameter_group_name       = aws_elasticache_parameter_group.main.id
+  port                       = 6379
+  security_group_ids         = [aws_security_group.redis.id]
+  subnet_group_name          = var.redis_subnet_group_name
+}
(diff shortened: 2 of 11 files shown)


terraform/modules/elasticache/parameter_groups.tf
```diff
@@ -0,0 +1,11 @@
+resource "aws_elasticache_parameter_group" "main" {
+  name = "${var.prefix}-redis-parameter-group"
+  # Ensure the redis version below tallies with the engine_version defined for the redis elasticache cluster (see redis.tf)
+  # Parameter group families are outlined here - https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/ParameterGroups.Redis.html
+  family = "redis6.x"
+
+  parameter {
+    name  = "maxmemory-policy"
```
Chirag-Bhatti, line 8: Observed this was set in PaaS so have carried it over to AWS

terraform/modules/elasticache/redis.tf
```diff
@@ -0,0 +1,18 @@
+#tfsec:ignore:aws-elasticache-enable-backup-retention:TODO CLDC-2679 setup a snapshot retention limit
+resource "aws_elasticache_cluster" "main" {
+  #checkov:skip=CKV_AWS_134:TODO CLDC-2679 setup a snapshot retention limit
+  #TODO CLDC-2679 setup redis replicas with multi-az and automatic failover
+  #TODO CLDC-2682 setup redis logging
+  cluster_id                 = "${var.prefix}-redis"
+  auto_minor_version_upgrade = true
+  apply_immediately          = true
+  engine                     = "redis"
+  engine_version             = "6.2"
```
Chirag-Bhatti, line 10: We use `6.2.6` in PaaS (Terraform here will just take the latest version i.e. `6.2.x`). You can also provide `6.x` which will get the latest minor version, however I'm not sure we'd want to do this in case of compatibility issues

terraform/production/main.tf
```diff
@@ -26,27 +26,41 @@ provider "aws" {
...
+  prefix                                  = local.prefix
+  vpc_cidr_block                          = "10.0.0.0/16"
+  vpc_flow_cloudwatch_log_expiration_days = 90
+}
+
+module "redis" {
+  source = "../modules/elasticache"
+
+  prefix                  = local.prefix
+  node_type               = "cache.t2.micro"
```
Chirag-Bhatti, line 56: PaaS uses `micro-6.x` across all envs, which looks roughly comparable to `cache.t2.micro` in AWS so I've done for all envs here

terraform/modules/elasticache/redis.tf
```diff
@@ -0,0 +1,18 @@
+#tfsec:ignore:aws-elasticache-enable-backup-retention:TODO CLDC-2679 setup a snapshot retention limit
+resource "aws_elasticache_cluster" "main" {
+  #checkov:skip=CKV_AWS_134:TODO CLDC-2679 setup a snapshot retention limit
+  #TODO CLDC-2679 setup redis replicas with multi-az and automatic failover
+  #TODO CLDC-2682 setup redis logging
```
Chirag-Bhatti, line 5: You might have already seen, but I've created tickets to setup Redis snapshots / availability, and Redis logging later on

terraform/modules/elasticache/redis.tf
```diff
@@ -0,0 +1,18 @@
+#tfsec:ignore:aws-elasticache-enable-backup-retention:TODO CLDC-2679 setup a snapshot retention limit
+resource "aws_elasticache_cluster" "main" {
+  #checkov:skip=CKV_AWS_134:TODO CLDC-2679 setup a snapshot retention limit
+  #TODO CLDC-2679 setup redis replicas with multi-az and automatic failover
+  #TODO CLDC-2682 setup redis logging
+  cluster_id                 = "${var.prefix}-redis"
+  auto_minor_version_upgrade = true
+  apply_immediately          = true
```
RachaelBooth, line 8: Do we want apply_immediately to be generally on? (I suspect that at the moment while we're likely to be changing things we do, but later we may not).

terraform/modules/elasticache/security_groups.tf
```diff
@@ -0,0 +1,25 @@
...
+  name        = "${var.prefix}-redis-security-group"
+  description = "Redis security group"
+  vpc_id      = var.vpc_id
+
+  ingress {
+    description = "Allow ingress on port 6379 from anything within the private subnet cidr block range"
+    protocol    = "tcp"
+    from_port   = 6379
+    to_port     = 6379
+    cidr_blocks = [var.private_subnet_cidr]
```
RachaelBooth, line 11: This might be better restricting on security group than on anything in our private subnet? (Although both would prevent external access which is the largest concern, I would normally expect restrictions to be done by security group except when being explicitly unrestrictive)

terraform/modules/elasticache/security_groups.tf
```diff
@@ -0,0 +1,25 @@
+resource "aws_security_group" "redis" {
+  name        = "${var.prefix}-redis-security-group"
+  description = "Redis security group"
+  vpc_id      = var.vpc_id
+
+  ingress {
```
RachaelBooth, line 6: It looks like the terraform docs have a note suggesting that people use `aws_vpc_security_group_ingress_rule` (and `..._egress_rule`) for new resources (sorry, I didn't end up looking at the docs for previous reviews). It's unclear whether we'll come up against they problem they mention, but it might be worth updating everywhere while that's still easy to change. 

(Although I'm saying that expecting it to be a quick thing to change, so let me know if that's wrong because we could decide to leave it until we actually hit a problem)

Chirag-Bhatti: Ah yes good spot! It doesn't look too bad to change (I've worked out the syntax already), but can't apply it because it doesn't let you get away with providing a blank security group source id. We don't have one until we make our ECS security group in a later PR, so I'd prefer to make the change there if that's ok (for RDS, DB and ECS sg's)?

terraform/modules/elasticache/security_groups.tf
Chirag-Bhatti: I agree - now updated 👍

terraform/modules/elasticache/redis.tf
Chirag-Bhatti: Indeed for now we do, later probably not (atleast not for prod). I've made a ticket to make sure we consider and change this down the line. See [CLDC-2687](https://dluhcdigital.atlassian.net/browse/CLDC-2687)

terraform/modules/elasticache/security_groups.tf
```diff
@@ -0,0 +1,25 @@
...
+    to_port         = 6379
+    security_groups = var.ingress_source_security_group_ids
+  }
+
+  egress {
+    description = "Allow egress from port 6379 to anything within the private subnet cidr block range"
+    protocol    = "tcp"
+    from_port   = 6379
+    to_port     = 6379
+    cidr_blocks = [var.private_subnet_cidr]
```
RachaelBooth, line 19: I think this could be by security group as well, for consistency (although it's less odd to be a range on the way out, we're not expecting this to need egress to the whole vpc).

RachaelBooth (approved):
Happy for updates to wait for a future PR as it sounds like that'll be easier.

PR #7, opened 2023-08-17
CLDC-2526: Setup ECS Cluster

Chirag-Bhatti:
- Setup ECS cluster with service and task definitions
- Add task definitions for running the app, db:setup, db:seed and db:migrate
- Update how we define security groups (using separate ingress / egress rule resources)
- Implement a parameter store and env vars for running the app
- Setup up <env>.auto.tfvars files so thay can be passed in to the parameter store (files not committed due to sensitivity)


terraform/modules/ecs/logging.tf
```diff
@@ -0,0 +1,12 @@
+# TODO CLDC-2542 remove/modify this temporary logging setup
+#tfsec:ignore:aws-cloudwatch-log-group-customer-key:TODO CLDC-2542 create and encrypt with a KMS key if retaining this log group
+resource "aws_cloudwatch_log_group" "main" {
```
Chirag-Bhatti, line 3: Unfortunately to view logs from ecs (useful for debugging issues with the app running on it), we have to setup some cloudwatch, and then provide this to `logConfiguration` in the `aws_ecs_task_definition`

terraform/development/main.tf
```diff
@@ -33,12 +33,39 @@ locals {
...
+
+module "ecs" {
+  source = "../modules/ecs"
+
+  prefix = local.prefix
+  additional_task_role_policy_arns = {
+    "RDS_access" : module.database.rds_data_access_policy_arn
+    "Redis_access" : "arn:aws:iam::aws:policy/AmazonElastiCacheFullAccess"
+  }
+  ecr_repository_url                = "815624722760.dkr.ecr.eu-west-2.amazonaws.com/core-ecr"
```
Chirag-Bhatti, line 52: I don't think I can output this from meta, so had to hardcode for now. Not found a better way yet (other than tucking it away in `tfvars` or similar)

terraform/development/main.tf
```diff
@@ -33,12 +33,39 @@ locals {
...
+  egress_to_db_security_group_id    = module.database.rds_security_group_id
+  egress_to_redis_security_group_id = module.redis.redis_security_group_id
+  ecs_environment_variables = [
+    { Name = "API_USER", Value = "dluhc-user" },
+    { Name = "APP_HOST", Value = "" },
+    { Name = "CSV_DOWNLOAD_PAAS_INSTANCE", Value = "" },
```
Chirag-Bhatti, line 58: Placeholding values for now

terraform/development/main.tf
```diff
@@ -49,19 +76,60 @@ module "networking" {
   vpc_flow_cloudwatch_log_expiration_days = 90
 }
 
-module "redis" {
-  source = "../modules/elasticache"
+module "parameter_store" {
```
Chirag-Bhatti, line 79: I'm not a fan of having to list all the params out like this (just makes this file really long), but not thought of a better way yet (prioritised getting ECS working)

terraform/modules/ecs/ecs.tf
```diff
@@ -0,0 +1,231 @@
...
+    }
+  ])
+
+  runtime_platform {
+    operating_system_family = "LINUX"
+    cpu_architecture        = "X86_64"
+  }
+}
+
+resource "aws_ecs_task_definition" "db_migrate" {
```
Chirag-Bhatti, line 57: Quite a bit of code duplication for each one of these task definitions. Mostly the same other than terraform name, `family` and `command` attributes

terraform/modules/ecs/ecs.tf
```diff
@@ -0,0 +1,231 @@
...
+    }
+  ])
+
+  runtime_platform {
+    operating_system_family = "LINUX"
+    cpu_architecture        = "X86_64"
+  }
+}
+
+resource "aws_ecs_task_definition" "db_setup" {
```
Chirag-Bhatti, line 159: One of the reasons the app wasn't running at first was because we have to setup the db first. So I ran this task definition on the ECS, and then was able to run the app. So I've added this here so we can reuse it in future (e.g when using ECS on dev and prod for the first time)

terraform/modules/ecs/security_groups.tf
```diff
@@ -0,0 +1,50 @@
...
+}
+
+resource "aws_vpc_security_group_egress_rule" "ecs_https_egress" {
+  description       = "Allow egress from port 443 to any IP address"
+  cidr_ipv4         = "0.0.0.0/0"
+  ip_protocol       = "tcp"
+  from_port         = 443
+  to_port           = 443
+  security_group_id = aws_security_group.ecs.id
+}
```
Chirag-Bhatti, line 50: It looks some or all of this is necessary otherwise the container can't get params from the param store! Might be able to narrow down the cidr block to just the NAT gateway IP range but not tried this yet

terraform/modules/elasticache/outputs.tf
```diff
@@ -0,0 +1,10 @@
+output "redis_connection_string" {
+  value       = "redis://${aws_elasticache_cluster.main.cache_nodes[0].address}:6379"
```
Chirag-Bhatti, line 2: I think I'll parameterise port numbers to avoid any issues were they ever to be changed and forgotten to be updated places like here

terraform/modules/ecs/ecs.tf
Chirag-Bhatti: An alternative would be to have just one task definition (for running the app); then whenever we want to manually run something like `db:migrate`, we run the same task definition but overwrite the `command` in the aws console

terraform/development/main.tf
```diff
@@ -33,12 +33,39 @@ locals {
...
+  vpc_id                             = module.networking.vpc_id
+}
+
+module "ecs" {
+  source = "../modules/ecs"
+
+  prefix = local.prefix
+  additional_task_role_policy_arns = {
+    "RDS_access" : module.database.rds_data_access_policy_arn
+    "Redis_access" : "arn:aws:iam::aws:policy/AmazonElastiCacheFullAccess"
```
Chirag-Bhatti, line 50: It feels like we might not be able to control who can read/write to Redis using an IAM policy (for example none of the elasticache actions listed in AWS console seem to be what we'd configure - see image)

It looks like this might need to be configured in Redis itself using an ACL ([see here](https://redis.io/docs/management/security/acl/) for example)

For now I've left this in, and down the line we'll need to check if it's actually needed to use Redis with the app

image

terraform/development/main.tf
```diff
@@ -33,12 +33,39 @@ locals {
...
+    { Name = "APP_HOST", Value = "" },
+    { Name = "CSV_DOWNLOAD_PAAS_INSTANCE", Value = "" },
+    { Name = "RAILS_ENV", Value = "staging" },
+    { Name = "RAILS_LOG_TO_STDOUT", Value = "true" },
+    { Name = "RAILS_SERVE_STATIC_FILES", Value = "true" }
+  ]
+  ecs_parameters         = module.parameter_store.parameter_arns
+  ecs_task_cpu           = 512
+  ecs_task_desired_count = 2
+  ecs_task_memory        = 1024
```
Chirag-Bhatti, line 66: In PaaS the application instances have the following specs:

Dev
- 1 instance with:
- Unknown CPU
- 1GiB RAM
- 1GiB Disk

Staging
- 2 instances each with:
- Unknown CPU
- 1GiB RAM
- 1GiB Disk

Production
- 4 instances each with:
- Unknown CPU
- Memory 1GiB
- Disk 1GiB

For now I've just configured all envs with the same but we should fine tune this down the line. Generally for staging and prod I'd expect to have at least 2 parallel tasks running

terraform/development/main.tf
Chirag-Bhatti: Since refactoring this is no longer valid - resolving this 😀

terraform/modules/elasticache/outputs.tf
Chirag-Bhatti: Port numbers have been made into variables now

terraform/modules/application/security_groups.tf
```diff
@@ -0,0 +1,54 @@
...
+resource "aws_vpc_security_group_ingress_rule" "ecs_ingress" {
+  description       = "Allow ingress on port ${var.application_port} from any IP address"
+  cidr_ipv4         = "0.0.0.0/0"
+  ip_protocol       = "tcp"
+  from_port         = var.application_port
+  to_port           = var.application_port
+  security_group_id = aws_security_group.ecs.id
+}
+
+resource "aws_vpc_security_group_egress_rule" "ecs_db_egress" {
```
RachaelBooth, line 20: I am surprised by these two, particularly by the port specification on the ECS side (although I'm not an expert on networking) - I'd expect this to be covered by the more generic egress rules.

terraform/modules/application/security_groups.tf
```diff
@@ -0,0 +1,54 @@
...
+  name        = "${var.prefix}-ecs-security_group"
+  description = "ECS security group"
+  vpc_id      = var.vpc_id
+
+  lifecycle {
+    create_before_destroy = true
+  }
+}
+
+resource "aws_vpc_security_group_ingress_rule" "ecs_ingress" {
```
RachaelBooth, line 11: N.B. Fine for the scope of this ticket, but should restrict this later once the load balancer is in play.

SamSeed-Softwire: Created ticket for this: https://dluhcdigital.atlassian.net/browse/CLDC-2710

terraform/modules/elasticache/variables.tf
```diff
@@ -1,6 +1,11 @@
-variable "ingress_source_security_group_ids" {
-  type        = list(string)
-  description = "The security group ids (sources) the redis security group will allow ingress from"
+variable "ingress_from_ecs_security_group_id" {
+  type        = string
+  description = "The id of the ecs security group for redis ingress"
+}
+
+variable "egress_to_ecs_security_group_id" {
```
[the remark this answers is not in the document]
SamSeed-Softwire: Have simplified to only use one variable (and haves similarly renamed the variable for the rds module too).

SamSeed-Softwire (approved):

terraform/modules/application/security_groups.tf
RachaelBooth: So I've had more of a think - the reason I'm surprised is that in applications I've worked with before we've not restricted outbound ports (because we trust that connections initiated by our app are fine), so a generic rule has covered it. 

Restricting is not unreasonable so long as we do make sure we've covered everything.

terraform/modules/application/ecs.tf
```diff
@@ -0,0 +1,295 @@
+#tfsec:ignore:aws-ecs-enable-container-insight:TODO CLDC-2542 enable container insights if necessary for logging/monitoring
+resource "aws_ecs_cluster" "main" {
+  #checkov:skip=CKV_AWS_65:TODO CLDC-2542 enable container insights if necessary for logging/monitoring
+  name = "${var.prefix}-ecs-cluster"
```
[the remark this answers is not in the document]
RachaelBooth: I've upgraded the priority of that ticket and cheekily pulled it into sprint, because it's many hours of work easier to change now than it is once we've got anything we can't destroy (e.g. because dns records are pointing at it).

Yeah, I should have caught it earlier but I do think that our convention should not include the resource type in names.

terraform/modules/application/security_groups.tf
SamSeed-Softwire: Agreed: happy to restrict.

terraform/modules/application/ecs.tf
```diff
@@ -0,0 +1,295 @@
...
+        { Name = "API_USER", Value = "dluhc-user" },
+        { Name = "APP_HOST", Value = var.app_host },
+        { Name = "CSV_DOWNLOAD_PAAS_INSTANCE", Value = "" },
+        { Name = "EXPORT_PAAS_INSTANCE", Value = "" },
+        { Name = "IMPORT_PAAS_INSTANCE", Value = "" },
+        { Name = "RAILS_ENV", Value = var.rails_env },
+        { Name = "RAILS_LOG_TO_STDOUT", Value = "true" },
+        { Name = "RAILS_SERVE_STATIC_FILES", Value = "true" },
+        { Name = "REDIS_INSTANCE_NAME", Value = "" },
+        { Name = "REDIS_CONFIG", Value = "[{\"instance_name\":\"\",\"credentials\":{\"uri\":\"${var.redis_connection_string}\"}}]" },
```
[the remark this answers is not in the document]
SamSeed-Softwire: Ticket raised for this: https://dluhcdigital.atlassian.net/browse/CLDC-2721.

PR #9, opened 2023-08-24
CLDC-2525: Set up load balancer

SamSeed-Softwire:
Create load balancer with permissive ingress/egress rules (we can tighten these up later if appropriate of course)

### DIFF
--- a/terraform/modules/front_door/security_groups.tf
+++ b/terraform/modules/front_door/security_groups.tf
@@ -9,6 +9,7 @@ resource "aws_security_group" "load_balancer" {
 }
 
 resource "aws_vpc_security_group_ingress_rule" "load_balancer_http_ingress" {
+  #checkov:skip=CKV_AWS_260:ingress from all IPs to port 80 required as load balancer is public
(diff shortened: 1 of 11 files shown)


terraform/modules/front_door/security_groups.tf
```diff
@@ -0,0 +1,55 @@
...
+resource "aws_vpc_security_group_ingress_rule" "load_balancer_https_ingress" {
+  description       = "Allow ingress on port 443 from all IP addresses"
+  cidr_ipv4         = "0.0.0.0/0"
+  ip_protocol       = "tcp"
+  from_port         = 443
+  to_port           = 443
+  security_group_id = aws_security_group.load_balancer.id
+}
+
+resource "aws_vpc_security_group_egress_rule" "load_balancer_http_egress" {
```
RachaelBooth, line 30: These two should actually not be necessary - what AWS suggests for an application load balancer outbound is just the instance security group on the instance listener port and health check port (if different).

Note that responses to inbound traffic should always be allowed, regardless of the outbound rules.

RachaelBooth (approved):

PR #10, opened 2023-08-25
CLDC-2701: Add xml export bucket

RachaelBooth:


RachaelBooth:
I have tested that this works on AWS (with updated application code, see https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1872 but also not requiring secure cookies)

PR #13, opened 2023-08-29
CLDC-2596: Add bulk upload bucket

SamSeed-Softwire:
This PR is almost identical to https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data-infrastructure/pull/10. The differences are:
- different names (`bulk_upload` rather than `export`)
- slightly updated checkov/tfsec comments

### DIFF
--- a/terraform/development/main.tf
+++ b/terraform/development/main.tf
@@ -60,28 +60,28 @@ module "cds_export" {
 module "application" {
   source = "../modules/application"
 
-  prefix                          = local.prefix
-  app_host                        = ""
-  application_port                = local.application_port
+  prefix                               = local.prefix
+  app_host                             = ""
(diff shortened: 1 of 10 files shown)


SamSeed-Softwire:
I have tested bulk upload on the staging app and found that it does indeed create files in the new bucket. I didn't test whether the app could read from the bucket correctly, but I can't see why it wouldn't given we can communicate with the bucket to put objects in it.

I'm not sure whether the app is currently communicating with Notify correctly, but that's not within the scope of this ticket.

RachaelBooth (approved):
Looks fine to me.

I realise the lifecycle config numbers are presumably coming from those for the cds export bucket which were pretty much plucked from thin air, but they seem sensible enough and we can change them later if we need to.

PR #16, opened 2023-08-31
CLDC-2545: set up CloudFront

SamSeed-Softwire:
Basic cloudfront distribution with minimal encryption settings implemented (this will be done in later tickets).

Note that there doesn't seem to be good Terraform documentation for all the config options. Instead I used the CloudFront API documentation to understand what all the options were: https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_Types.html.

### DIFF
--- a/terraform/modules/front_door/cloudfront.tf
+++ b/terraform/modules/front_door/cloudfront.tf
@@ -1,0 +1,47 @@
+locals {
+  origin_id       = "${var.prefix}-origin"
+}
+
+resource "aws_cloudfront_distribution" "this" {
+  enabled = true
+  is_ipv6_enabled = true
+  price_class     = "PriceClass_100" # Affects which edge locations are used by cloudfront, which affects the latency users will experience in different geographic areas
+
+  origin {
+    domain_name = aws_lb.main.dns_name
+    origin_id   = local.origin_id
+
+    custom_origin_config {
+      http_port                = 80
+      https_port               = 443
+      origin_keepalive_timeout = 60
+      origin_protocol_policy   = "match-viewer"
+      origin_read_timeout      = 60
+      origin_ssl_protocols     = ["TLSv1.2"]
+    }
+  }
+
+  default_cache_behavior {
+    allowed_methods          = ["DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"]
+    cached_methods           = ["GET", "HEAD", "OPTIONS"]
+    cache_policy_id          = aws_cloudfront_cache_policy.ttl_based.id
+    compress                 = true
+    origin_request_policy_id = aws_cloudfront_origin_request_policy.this.id
+    target_origin_id         = local.origin_id
+    viewer_protocol_policy   = "allow-all"
+  }
+
+  viewer_certificate {
+    cloudfront_default_certificate = true
+  }
+
+  restrictions {
+    geo_restriction {
+      restriction_type = "none"
+    }
+  }
+
+  tags = {
+    Name = "${var.prefix}-cloudfront"
+  }
+}
--- a/terraform/modules/front_door/cloudfront.tf
+++ b/terraform/modules/front_door/cloudfront.tf
@@ -14,6 +14,7 @@ resource "aws_cloudfront_distribution" "this" {
   #checkov:skip=CKV_AWS_86:TODO we will be implementing logging later
   #checkov:skip=CKV_AWS_174:TODO CLDC-2680
   #checkov:skip=CKV_AWS_305:no need to define a default root object because the root of our distribution is just the app's homepage
+  #checkov:skip=CKV_AWS_310:we have decided that we're unlikely to need a secondary load balancer
   enabled         = true
   is_ipv6_enabled = true
   price_class     = "PriceClass_100" # Affects which edge locations are used by cloudfront, which affects the latency users will experience in different geographic areas
--- a/terraform/modules/front_door/cloudfront.tf
+++ b/terraform/modules/front_door/cloudfront.tf
@@ -2,7 +2,18 @@ locals {
   origin_id       = "${var.prefix}-origin"
 }
 
+#tfsec:ignore:aws-cloudfront-enable-logging:TODO we will be implementing logging later
+#tfsec:ignore:aws-cloudfront-enable-waf:TODO CLDC-2546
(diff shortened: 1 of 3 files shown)


terraform/modules/front_door/cloudfront.tf
```diff
@@ -0,0 +1,58 @@
...
+  price_class     = "PriceClass_100" # Affects which edge locations are used by cloudfront, which affects the latency users will experience in different geographic areas
+
+  origin {
+    domain_name = aws_lb.main.dns_name
+    origin_id   = local.origin_id
+
+    custom_origin_config {
+      http_port                = 80
+      https_port               = 443
+      origin_keepalive_timeout = 60
```
SamSeed-Softwire, line 28: Is this necessary? It's what was used on EPB.

terraform/modules/front_door/cloudfront.tf
```diff
@@ -0,0 +1,58 @@
...
+      origin_protocol_policy   = "match-viewer"
+      origin_read_timeout      = 60
```
SamSeed-Softwire, line 30: Is this necessary? It's what was used on EPB.

terraform/modules/front_door/cloudfront.tf
```diff
@@ -0,0 +1,58 @@
...
+
+  origin {
+    domain_name = aws_lb.main.dns_name
+    origin_id   = local.origin_id
+
+    custom_origin_config {
+      http_port                = 80
+      https_port               = 443
+      origin_keepalive_timeout = 60
+      origin_protocol_policy   = "match-viewer"
```
SamSeed-Softwire, line 29: Should this be "http-only" for now?

terraform/modules/front_door/cloudfront.tf
```diff
@@ -0,0 +1,58 @@
...
+    cached_methods           = ["GET", "HEAD", "OPTIONS"]
+    cache_policy_id          = aws_cloudfront_cache_policy.ttl_based.id
+    compress                 = true
+    origin_request_policy_id = aws_cloudfront_origin_request_policy.this.id
+    target_origin_id         = local.origin_id
+    viewer_protocol_policy   = "allow-all"
+  }
+
+  viewer_certificate {
+    cloudfront_default_certificate = true
```
SamSeed-Softwire, line 46: "If the distribution doesn't use Aliases (also known as alternate domain names or CNAMEs)—that is, if the distribution uses the CloudFront domain name such as d111111abcdef8.cloudfront.net—**set CloudFrontDefaultCertificate to true and leave all other fields empty**."

Source: top of https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_ViewerCertificate.html.

terraform/modules/front_door/cloudfront.tf
```diff
@@ -0,0 +1,58 @@
...
+#tfsec:ignore:aws-cloudfront-enforce-https:TODO CLDC-2654
+#tfsec:ignore:aws-cloudfront-use-secure-tls-policy:TODO CLDC-2680
+resource "aws_cloudfront_distribution" "this" {
+  #checkov:skip=CKV_AWS_34:TODO CLDC-2654
+  #checkov:skip=CKV2_AWS_42:TODO CLDC-2680
+  #checkov:skip=CKV2_AWS_47:TODO CLDC-2546 when setting up WAF it should be configured appropriately to mitigate against the Log4j vulnerability https://docs.bridgecrew.io/docs/ensure-aws-cloudfront-attached-wafv2-webacl-is-configured-with-amr-for-log4j-vulnerability
+  #checkov:skip=CKV_AWS_68:TODO CLDC-2546
+  #checkov:skip=CKV_AWS_86:TODO we will be implementing logging later
+  #checkov:skip=CKV_AWS_174:TODO CLDC-2680
+  #checkov:skip=CKV_AWS_305:no need to define a default root object because the root of our distribution is just the app's homepage
```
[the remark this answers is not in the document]
RachaelBooth: I think we don't want multiple origins - we'll have various failover things built in in the ECS service, but not expecting to have secondary load balancer at the moment.

I suspect we do want a response headers policy - looking at the managed ones I think we definitely want one that includes the security headers, not immediately sure if we need cors but actually probably not?

terraform/modules/front_door/cloudfront.tf
```diff
@@ -0,0 +1,58 @@
...
+  enabled         = true
+  is_ipv6_enabled = true
```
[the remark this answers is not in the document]
RachaelBooth: Maybe, but the default doesn't look unreasonable so happy to leave it.

terraform/modules/front_door/cloudfront.tf
RachaelBooth: I guess so - our load balancer can't accept https yet.

terraform/modules/front_door/cloudfront.tf
RachaelBooth: Hmm, unclear what is a reasonable timeout to have for our app. I'd be tempted to stick with 30 (which I think is the default) unless we see problems.

terraform/modules/front_door/policies.tf
```diff
@@ -0,0 +1,38 @@
+resource "aws_cloudfront_cache_policy" "ttl_based" {
```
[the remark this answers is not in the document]
RachaelBooth: I think this behaviour looks alright.

terraform/modules/front_door/cloudfront.tf
SamSeed-Softwire: The default keepalive timeout is 5s (min 1s, max 60s). The default read timeout is 30s (min 1s, max 60s). Let's discuss in a bit

terraform/modules/front_door/cloudfront.tf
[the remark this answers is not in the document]
SamSeed-Softwire: Agreed to go with `http2and3`.

terraform/modules/front_door/cloudfront.tf
SamSeed-Softwire: Agreed to go with the default.

terraform/modules/front_door/cloudfront.tf
SamSeed-Softwire: Agreed to go with the default.

PR #26, opened 2023-09-07
CLDC-2680: Cloudfront certificate + custom URL

SamSeed-Softwire:
Have only updated the staging entrypoint for now - once we've agreed on what it should look like I'll replicate it for dev and prod.

Will `fmt` once everything else has been finalised.

### DIFF
--- a/terraform/modules/certificates/outputs.tf
+++ b/terraform/modules/certificates/outputs.tf
@@ -1,4 +1,9 @@
-output "certificate_arn" {
-  value       = aws_acm_certificate.this.arn
-  description = "The arn of the certificate"
+output "cloudfront_certificate_arn" {
+  value       = aws_acm_certificate.cloudfront.arn
+  description = "The arn of the cloudfront certificate"
+}
+
+output "cloudfront_certificate_validation" {
+  value       = aws_acm_certificate.cloudfront.domain_validation_options
+  description = "The domain validation objects for the cloudfront certificate"
 }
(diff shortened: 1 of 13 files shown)


terraform/modules/certificates/variables.tf
```diff
@@ -0,0 +1,4 @@
+variable "domain_name" {
+  type        = string
+  description = "Then domain name for which the certificate should be issued"
```
RachaelBooth, line 3: I think it would be better to make this module create all the necessary certs for a given environment, rather than just be a wrapper for "make a certificate". Which means adjusting the inputs and naming to be sensible for that case.

terraform/modules/certificates/outputs.tf
```diff
@@ -0,0 +1,4 @@
+output "certificate_arn" {
+  value       = aws_acm_certificate.this.arn
+  description = "The arn of the certificate"
+}
```
RachaelBooth, line 4: Is there a way we can output what validation records need to be added (rather than needing to go and look them up)?

(N.B. This is a nice to have - only worth a small amount of time)

terraform/modules/front_door/variables.tf
```diff
@@ -1,8 +1,18 @@
+variable "app_host" {
+  type        = string
+  description = "The value of the app host environment variable"
```
RachaelBooth, line 3: This description needs updating to make sense here.

RachaelBooth (approved):
Looks good to me, pending the validation check / formatting.

You probably want to output the required validation records from the staging module itself (otherwise they'll still be tricky to find).

PR #28, opened 2023-09-08
CLDC-2654: HTTPS between cloudfront and load balancer

SamSeed-Softwire:
Will `fmt` once everything else is done.

### DIFF
--- a/terraform/modules/front_door/load_balancer.tf
+++ b/terraform/modules/front_door/load_balancer.tf
@@ -1,7 +1,6 @@
 #tfsec:ignore:aws-elb-alb-not-public:load balancer is exposed to internet as it receives traffic from public
 resource "aws_lb" "this" {
   #checkov:skip=CKV_AWS_91:setup access logs on load balancer TODO CLDC-2705
-  #checkov:skip=CKV2_AWS_20:redirect http requests to https TODO CLDC-2654
   #checkov:skip=CKV2_AWS_28:WAF protection to be setup TODO CLDC-2546
   name                       = var.prefix
   drop_invalid_header_fields = true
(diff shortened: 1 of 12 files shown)


terraform/modules/front_door/load_balancer.tf
```diff
@@ -15,28 +14,27 @@ resource "aws_lb" "this" {
 resource "aws_lb_target_group" "this" {
   name        = var.prefix
   port        = var.application_port
-  protocol    = "HTTP"
+  protocol    = "HTTPS"
   vpc_id      = var.vpc_id
   target_type = "ip"
```
SamSeed-Softwire, line 19: The target type is "ip" but we're not actually directly specifying any IP addresses. Do we not want a [`aws_lb_target_group_attachment`](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data-infrastructure/pull/28/files) block to associate it with the container? Or is this redundant because we've defined a `load_balancer` configuration block in the ECS service definition?

terraform/staging/main.tf
```diff
@@ -40,6 +40,7 @@ locals {
   app_host         = "staging.submit-social-housing-data.levellingup.gov.uk"
   application_port = 8080
   database_port    = 5432
+  load_balancer_domain_name = "staginglb.submit-social-housing-data.levellingup.gov.uk"
```
RachaelBooth, line 43: Can I suggest this is `staging.lb.submit-social-housing-data.levellingup.gov.uk`? (It doesn't really make a difference, just is a nicer pattern of making it always `staging.whatever-prod-is`)

terraform/modules/front_door/load_balancer.tf
RachaelBooth: It's not necessary in this setup - we associate the target group with our service in the ECS service definition. 

Also I think the target group attachment resource would only allow associating with a specific instance or container id, which we don't want.

RachaelBooth (approved):
One small domain name comment, but otherwise looks good to me.

PR #30, opened 2023-09-11
CLDC-2661: RDS backups / restorability

SamSeed-Softwire:
https://dluhcdigital.atlassian.net/browse/CLDC-2661

- Will duplicate code for dev and prod, and also`fmt`, when everything else on ticket is done.

### DIFF
--- a/terraform/modules/rds/database.tf
+++ b/terraform/modules/rds/database.tf
@@ -1,4 +1,3 @@
-#tfsec:ignore:aws-rds-specify-backup-retention:TODO CLDC-2661
 #tfsec:ignore:aws-rds-enable-performance-insights:TODO CLDC-2660 if necessary
 #tfsec:ignore:AVD-AWS-0176:iam authentication not suitable as tokens only last 15minutes, password authentication preferred
 resource "aws_db_instance" "this" {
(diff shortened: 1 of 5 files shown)


terraform/modules/rds/database.tf
```diff
@@ -8,21 +7,27 @@ resource "aws_db_instance" "this" {
   #checkov:skip=CKV_AWS_353:performance insights TODO CLDC-2660 if necessary
   #checkov:skip=CKV_AWS_354:performance insights TODO CLDC-2660 if insights are necessary
   #checkov:skip=CKV2_AWS_30:query logging TODO CLDC-2660
-  #checkov:skip=CKV2_AWS_60:copy tags to snapshots TODO CLDC-2661
   identifier                 = var.prefix
   apply_immediately          = true
   auto_minor_version_upgrade = true
   allocated_storage          = var.allocated_storage #units are GiB
+  backup_retention_period    = var.backup_retention_period
+  backup_window              = "23:09-23:39"
```
SamSeed-Softwire, line 15: We could also use a `[aws_backup_plan](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/backup_plan.html)` block, but I don't think it's necessary.

terraform/modules/rds/database.tf
```diff
@@ -8,21 +7,27 @@ resource "aws_db_instance" "this" {
...
+  copy_tags_to_snapshot      = true
```
SamSeed-Softwire, line 16: We don't have any tags currently, but it seems wise to want this in case we do ever have tags.

terraform/modules/rds/database.tf
```diff
@@ -8,21 +7,27 @@ resource "aws_db_instance" "this" {
...
   db_name                    = "data_collector"
   db_subnet_group_name       = var.db_subnet_group_name
+  delete_automated_backups   = false
```
SamSeed-Softwire, line 19: This defaults to true but it seems wise to set this as false and then manually delete backups, as an extra layer of protection against data loss.

terraform/modules/rds/database.tf
```diff
@@ -8,21 +7,27 @@ resource "aws_db_instance" "this" {
...
   engine                     = "postgres"
   engine_version             = "13.11"
+  final_snapshot_identifier  = var.prefix
   instance_class             = var.instance_class
+  maintenance_window         = "Mon:02:33-Mon:03:03"
   multi_az                   = true
   password                   = random_password.this.result
   port                       = var.database_port
   publicly_accessible        = false
+  skip_final_snapshot        = false
```
SamSeed-Softwire, line 30: This defaults to false, but I think it's sensible to clearly show it's false here.

terraform/modules/rds/database.tf
SamSeed-Softwire, line 15: I didn't think we'd want different backup and maintenance windows for the different environments, so didn't make this into a variable.

terraform/modules/rds/database.tf
```diff
@@ -8,21 +7,27 @@ resource "aws_db_instance" "this" {
...
+  copy_tags_to_snapshot      = true
   db_name                    = "data_collector"
   db_subnet_group_name       = var.db_subnet_group_name
+  delete_automated_backups   = false
   deletion_protection        = true # needs to be set to false and applied if you need to delete the DB
   engine                     = "postgres"
   engine_version             = "13.11"
+  final_snapshot_identifier  = var.prefix
   instance_class             = var.instance_class
+  maintenance_window         = "Mon:02:33-Mon:03:03"
```
SamSeed-Softwire, line 25: I didn't think we'd want different backup and maintenance windows for the different environments, so didn't make this into a variable.

RachaelBooth (approved):
LGTM (subject to formatting and updating the dev and prod modules)

PR #31, opened 2023-09-11
CLDC-2546: Set up WAF

Chirag-Bhatti:
Setup WAF to cover our cloudfront distribution with the following AWS managed rules suggested on the ticket:
- `AWSManagedRulesCommonRuleSet` (excluding the `SizeRestrictions_BODY` rule to allow file uploads)
- `AWSManagedRulesKnownBadInputsRuleSet`
- `AWSManagedRulesAmazonIpReputationList`

I also added the following rule sets as they applied to our use case (as prompted by the ticket):
- `AWSManagedRulesSQLiRuleSet`
- `AWSManagedRulesLinuxRuleSet`
- `AWSManagedRulesUnixRuleSet`
See my comment on the rule priority order for more info about the rules and what they cover

### DIFF
--- a/terraform/modules/front_door/waf.tf
+++ b/terraform/modules/front_door/waf.tf
@@ -1,0 +1,164 @@
+provider "aws" {
+  alias  = "north-virginia"
+  region = "us-east-1"
+
+  assume_role {
+    role_arn = var.provider_role_arn
+  }
+}
+
+resource "aws_wafv2_web_acl" "this" {
+  #checkov:skip=CKV2_AWS_31:TODO CLDC-2781 setup WAF logging in cloudwatch
+  name        = var.prefix
+  description = "Web ACL to restrict traffic to CloudFront"
+  provider    = aws.north-virginia
+  scope       = "CLOUDFRONT"
+
+  default_action {
+    allow {}
+  }
+
+  visibility_config {
+    cloudwatch_metrics_enabled = true
+    metric_name                = "waf-metrics"
+    sampled_requests_enabled   = false
+  }
+
+  rule {
+    name     = "AWSManagedRulesAmazonIpReputationList"
+    priority = 1
+
+    override_action {
+      none {}
+    }
+
+    statement {
+      managed_rule_group_statement {
+        name        = "AWSManagedRulesAmazonIpReputationList"
+        vendor_name = "AWS"
+      }
+    }
+
+    visibility_config {
+      cloudwatch_metrics_enabled = true
+      metric_name                = "waf-block-disreputable-ip-metrics"
+      sampled_requests_enabled   = true
+    }
+  }
+
+  rule {
+    name     = "AWSManagedRulesCommonRuleSet"
+    priority = 2
+
+    override_action {
+      none {}
+    }
+
+    statement {
+      managed_rule_group_statement {
+        name        = "AWSManagedRulesCommonRuleSet"
+        vendor_name = "AWS"
+        rule_action_override {
+          name = "SizeRestrictions_BODY"
+          action_to_use {
+            allow {}
+          }
+        }
+      }
+    }
+
+    visibility_config {
+      cloudwatch_metrics_enabled = true
+      metric_name                = "waf-block-common-exploit-metrics"
+      sampled_requests_enabled   = true
+    }
+  }
+
+  rule {
+    name     = "AWSManagedRulesKnownBadInputsRuleSet"
+    priority = 3
+
+    override_action {
+      none {}
+    }
+
+    statement {
+      managed_rule_group_statement {
+        name        = "AWSManagedRulesKnownBadInputsRuleSet"
+        vendor_name = "AWS"
+      }
+    }
+
+    visibility_config {
+      cloudwatch_metrics_enabled = true
+      metric_name                = "waf-block-bad-input-exploit-metrics"
+      sampled_requests_enabled   = true
+    }
+  }
+
+  rule {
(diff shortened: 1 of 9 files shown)


terraform/modules/front_door/waf.tf
```diff
@@ -0,0 +1,164 @@
...
+    cloudwatch_metrics_enabled = true
+    metric_name                = "waf-metrics"
+    sampled_requests_enabled   = false
+  }
+
+  rule {
+    name     = "AWSManagedRulesAmazonIpReputationList"
+    priority = 1
+
+    override_action {
```
Chirag-Bhatti, line 31: Looks unnecessary but you have to provide this or an `action` block as noted [here](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_web_acl.html#rule-block)

terraform/modules/front_door/waf.tf
```diff
@@ -0,0 +1,164 @@
...
+    allow {}
+  }
+
+  visibility_config {
+    cloudwatch_metrics_enabled = true
+    metric_name                = "waf-metrics"
+    sampled_requests_enabled   = false
+  }
+
+  rule {
```
Chirag-Bhatti, line 27: Wasn't completely sure about rule order / priority here but generally felt this was the likely order of importance. I think as long as a bad request gets caught the priority doesn't matter too much:

- `AWSManagedRulesAmazonIpReputationList` - block IPs with a bad rep first
- `AWSManagedRulesCommonRuleSet` - block requests with common web app exploits
- `AWSManagedRulesKnownBadInputsRuleSet` - block requests with bad request patterns known for common exploits or discovery of vulnerabilities
- `AWSManagedRulesSQLiRuleSet` - block requests associated with exploitation of SQL databases
- `AWSManagedRulesLinuxRuleSet` - block requests associated with exploits specific to LINUX systems
- `AWSManagedRulesUnixRuleSet` - block requests associated with exploits specific to POSIX systems (AWS suggests this should be used with the Linux rule set)

More info about these can be found [here](https://docs.aws.amazon.com/waf/latest/developerguide/aws-managed-rule-groups-list.html). See the `baseline`, `use-case specific` and `IP reputation` sections

terraform/modules/front_door/waf.tf
```diff
@@ -0,0 +1,164 @@
...
+      managed_rule_group_statement {
+        name        = "AWSManagedRulesAmazonIpReputationList"
+        vendor_name = "AWS"
+      }
+    }
+
+    visibility_config {
+      cloudwatch_metrics_enabled = true
+      metric_name                = "waf-block-disreputable-ip-metrics"
+      sampled_requests_enabled   = true
```
Chirag-Bhatti, line 45: For all of our WAF rules, I thought it would be good to have sample requests enabled if a bad request is ever trapped by a rule. I believe these are free according to the console , see screenshot:
image

terraform/modules/front_door/waf.tf
```diff
@@ -0,0 +1,164 @@
+provider "aws" {
+  alias  = "north-virginia"
+  region = "us-east-1"
```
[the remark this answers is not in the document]
RachaelBooth: It's not (but probably should be) how we do it for the backend, but if you look at Sam's `certificates` branch you'll see an example of creating another provider in the entrypoint module and passing it in, which I think is generally better practice - and you can do this for passing in more than one aws provider with different keys to reference them.

It does mean doing that in each entrypoint module, but it means only the entrypoint modules are needing to care about the role that terraform is using etc.

terraform/modules/front_door/waf.tf
```diff
@@ -0,0 +1,164 @@
...
+    statement {
+      managed_rule_group_statement {
+        name        = "AWSManagedRulesAmazonIpReputationList"
+        vendor_name = "AWS"
+      }
+    }
+
+    visibility_config {
+      cloudwatch_metrics_enabled = true
+      metric_name                = "waf-block-disreputable-ip-metrics"
```
SamSeed-Softwire, line 44: Minor point - do we perhaps want to suffix individual metrics with `metric` rather than `metrics`? Or do we even need a suffix at all - might it not be obvious what it is when looking in the console?

terraform/modules/front_door/waf.tf
```diff
@@ -0,0 +1,164 @@
...
+  scope       = "CLOUDFRONT"
+
+  default_action {
+    allow {}
+  }
+
+  visibility_config {
+    cloudwatch_metrics_enabled = true
+    metric_name                = "waf-metrics"
+    sampled_requests_enabled   = false
```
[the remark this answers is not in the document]
Chirag-Bhatti: This is the top level visibility config, so I think will sample all requests coming into the WAF before exposed to the rules which we probably don't want / need

terraform/modules/front_door/waf.tf
Chirag-Bhatti: Yeah I think this was added in EPB to make the code more readable and I did something similar here, but in the console it's probably just noise (see screenshot), so I'll remove it 👍 

image

Chirag-Bhatti: Removed 👍

terraform/modules/front_door/waf.tf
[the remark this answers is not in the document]
Chirag-Bhatti: I've made this change for this module and the for the backend. Thanks for pointing it out! 😊

PR #35, opened 2023-09-12
Run Terraform checks on pushes to any branch

SamSeed-Softwire:


.github/workflows/terraform_checks.yml
```diff
@@ -7,8 +7,6 @@ on:
       - synchronize
       - reopened
   push:
```
Chirag-Bhatti, line 9: I suspect this works fine 😊. From a quick look I came across the below, which might be clearer and easier to adapt later down the line (while behaving the same)

```
on:
  push:
    branches:
      - '*'
```

Chirag-Bhatti: Happy with either if you don't feel like it's worth trying the above

RachaelBooth (approved):
LGTM

Should maybe consider ignoring dev at some point? (since that's going to get lots of messy commits where the results may well be someone else's problem, it seems like what you'll actually be using is the results for your feature branch)

PR #36, opened 2023-09-13
CLDC-2764: CDS access to export bucket

Chirag-Bhatti:
- Add a role for CDS to assume
- Add a policy allowing the role read access to the export bucket


terraform/development/main.tf
```diff
@@ -84,7 +84,8 @@ module "bulk_upload" {
 module "cds_export" {
   source = "../modules/cds_export"
 
-  prefix = local.prefix
+  prefix              = local.prefix
+  cds_access_role_arn = "arn:aws:iam::837698168072:role/developer" # TODO CLDC-2764 update role arn to that the CDS team will use
```
Chirag-Bhatti, line 88: Just placeholders for each env right now (terraform moans if left as an empty string)

terraform/modules/cds_export/policy.tf
```diff
@@ -0,0 +1,37 @@
...
+      "s3:Get*",
+      "s3:List*",
+      "s3:Describe*"
+    ]
+
+    effect = "Allow"
+
+    resources = [
+      aws_s3_bucket.this.arn,
+      "${aws_s3_bucket.this.arn}/*"
```
Chirag-Bhatti, line 29: I think this is necessary to allow access to the bucket and all the files at the root of it (i.e. the csv's and zip files)

RachaelBooth: That wildcard is, yes.

I think overall this policy can be more restrictive however - e.g. I don't think they need to be able to retrieve all of the bucket config information, just list the bucket contents and be able to retrieve any object in the bucket. (This probably involves two statements in this policy to separate actions allowed on the bucket from actions allowed on the contents of the bucket)

As it's read access it's not a huge problem, but it is worth looking briefly at restricting it further.

terraform/modules/cds_export/role_cds.tf
```diff
@@ -0,0 +1,9 @@
+resource "aws_iam_role" "cds" {
```
RachaelBooth, line 1: I think it's confusing to have this in a different file to the assume role policy document, and similarly for the policy attachment below, since almost always you want to see the associated policy content when you look at a role - can you combine them?

This would also prevent the "policy.tf" file naming from being confusing for not containing the "read and write objects" policy that we apply to the task role, although with the expansion of this module that section may want to be pulled out into it's own file now (which I'd think would be different from the file containing the cds bits) - maybe not worth it, I'll leave you to ponder that.

terraform/development/main.tf
RachaelBooth: Rather than providing something here that we don't want to apply the policy to, can we make this optional and have the module not set up the role/policy if it isn't provided?

(Particularly since long term I expect this not to be necessary for dev and staging)

terraform/modules/cds_export/role_cds.tf
Chirag-Bhatti: Sure I think combined makes more sense come to think of it as well. It's worth pulling out the "read and write" policy imo too, so I'll make those changes 👍

Chirag-Bhatti: Done 👍

terraform/development/main.tf
Chirag-Bhatti: Yeah good point! I've updated this now such that nothing will be setup for all envs as things stand, but can provide the cds role arn later for whichever env(s) needing it

terraform/modules/cds_export/policy.tf
Chirag-Bhatti: Think I've managed to resolve this if you can sanity check it 😊

RachaelBooth (approved):

PR #37, opened 2023-09-13
CLDC-2727: WAF rate limit

Chirag-Bhatti:
- Implement a WAF rule for rate limiting requests per IP over a 5min period (this is the default period)


terraform/modules/front_door/waf.tf
```diff
@@ -152,4 +152,26 @@ resource "aws_wafv2_web_acl" "this" {
...
+    priority = 7
+
+    action {
+      block {}
+    }
+
+    statement {
+      rate_based_statement {
+        aggregate_key_type = "IP"
+        limit              = 1000
```
Chirag-Bhatti, line 167: I'm not sure about a limit yet, but wanted to get the implementation reviewed while a determine a sensible threshold

Chirag-Bhatti: We may need to increase this limit during performance testing temporarily, I've added a note about this on the ticket [here](https://dluhcdigital.atlassian.net/browse/CLDC-2737). Hopefully not, but I recall on EPB we hit the limit because all our requests were being sent from a single IP during testing.

RachaelBooth: I think I was expecting this to only apply to the login path (actually that should probably be all paths you can reach without being logged in, so e.g. reset password as well) - which does most of the protection (assuming relative trust in our users), and allows us to set a much lower limit (because number of requests for legitimate use will be lower). 

Delta has an example of that that I was working from - they actually have an overall rate limit and a lower limit for login routes. 

I appreciate the ticket isn't particularly well laid out, and an overall limit is separately a good thing - in hindsight I'm actually fairly tempted to have both if you're up for implementing that as well?

SamSeed-Softwire (commented):
I have nothing to add!

terraform/modules/front_door/waf.tf
Chirag-Bhatti: Sorry - looks like I hadn't read the ticket properly! But yes up for implementing all of that 👍

terraform/modules/front_door/waf.tf
```diff
@@ -50,6 +50,11 @@ resource "aws_wafv2_web_acl" "this" {
         name        = "AWSManagedRulesCommonRuleSet"
         vendor_name = "AWS"
         rule_action_override {
+          # This rule blocks request bodies over 8KB in size, but CORE needs file uploads so we remove this restriction
```
Chirag-Bhatti, line 53: Thought this would be useful. Delta had a similar comment

terraform/modules/front_door/waf.tf
```diff
@@ -152,4 +157,97 @@ resource "aws_wafv2_web_acl" "this" {
...
+        custom_response {
+          response_code = 429
+        }
+      }
+    }
+
+    statement {
+      rate_based_statement {
+        aggregate_key_type = "IP"
+        limit              = 2000
```
Chirag-Bhatti, line 217: I think this a reasonable limit looking at prod on logit logs. Here's a profile of an IP address I found, that looked to be high load user pretty recently. They made a maximum of ~280 requests/min => ~1400 requests/5mins, so 2000 is a bit of headroom on this.

image

terraform/modules/front_door/waf.tf
```diff
@@ -152,4 +157,97 @@ resource "aws_wafv2_web_acl" "this" {
...
+    regex_string = "/account/sign-in"
+  }
+
+  regular_expression {
+    regex_string = "/account/two-factor-authentication"
+  }
+
+  regular_expression {
+    regex_string = "/account/two-factor-authentication/resend"
+  }
```
Chirag-Bhatti, line 252: Looking at `routes.rb` in the app codebase and having a play with the site, I think these are all the login routes that would need a separate "login" rate limit. AFAICT that covers everything, but pointing out as I'm not as familiar with the application.

terraform/modules/front_door/waf.tf
```diff
@@ -152,4 +157,97 @@ resource "aws_wafv2_web_acl" "this" {
...
+        custom_response {
+          response_code = 429
+        }
+      }
+    }
+
+    statement {
+      rate_based_statement {
+        aggregate_key_type = "IP"
+        limit              = 100
```
Chirag-Bhatti, line 176: This login IP rate limit is something Delta had and it looked reasonable to me. It's equivalent to 20 requests/min, which I think any IP doing above that is probably behaving badly

terraform/modules/front_door/waf.tf
Chirag-Bhatti: We could increase the limit to 64KB but this is not something Delta did. Depends on how much we want to look at file upload requests (presuming I've understood this correctly - that such requests are more risky because we can't check as much of the request body, and other request types will probably be below 16KB )

terraform/modules/front_door/waf.tf
```diff
@@ -152,4 +157,97 @@ resource "aws_wafv2_web_acl" "this" {
...
   }
+
+  rule {
+    name     = "login-ip-rate-limit"
+    priority = 7
+
+    action {
+      block {
+        custom_response {
+          response_code = 429
```
Chirag-Bhatti, line 168: I believe there is application logic that ties a 429 HTTP response to a "too many requests" error page, so thought it might be good to induce this on the two rate limiting rules we have here

terraform/modules/front_door/waf.tf
Chirag-Bhatti: Just to clarify, this was a high load user looking at all usage ever on prod

terraform/modules/front_door/waf.tf
Chirag-Bhatti: Having investigated further, it doesn't look like the application automatically redirects to this page if it sees a 429 response. Seems to be only if we're using rack attack for rate limiting, do you know if this is expected? Nonetheless I think it is worth still responding with a 429 (unless we don't obfuscate there is rate limiting from the client?)

Chirag-Bhatti:
I've also managed to test the rate limits function for login pages and overall scenarios, they seem to be working 😀
image

terraform/modules/front_door/waf.tf
RachaelBooth: So I think when rack attack rate limits (which we may only have set up on password reset requests?) it actually forwards to a "/429" route and we map that to an error page. 

If we return a 429 here it'll be before it hits the application, so I'm not expecting we can do nice in-app errors. We presumably could set this up to return a nicer thing, but not bothered about it at the moment - think just responding with a 429 will do.

terraform/modules/front_door/waf.tf
RachaelBooth: Yeah. Again, we'll want to check some realistic uses (but that can happen later).

RachaelBooth (approved):

terraform/modules/front_door/waf.tf
Chirag-Bhatti: Added a note about this in functionality regression testing ticket [CLDC-2738](https://dluhcdigital.atlassian.net/browse/CLDC-2738)

terraform/modules/front_door/waf.tf
Chirag-Bhatti: Added a note about this in functionality regression testing ticket [CLDC-2738](https://dluhcdigital.atlassian.net/browse/CLDC-2738)

PR #38, opened 2023-09-13
CLDC-2617: Ensure staging environment set up with appropriate sizing

SamSeed-Softwire:
Everything in this PR has already been discussed and agreed.

### DIFF
--- a/terraform/development/main.tf
+++ b/terraform/development/main.tf
@@ -127,7 +127,7 @@ module "redis" {
 
   prefix                  = local.prefix
   ecs_security_group_id   = module.application.ecs_security_group_id
-  node_type               = "cache.t2.micro"
+  node_type               = "cache.t4g.micro"
   redis_port              = local.redis_port
   redis_subnet_group_name = module.networking.redis_private_subnet_group_name
   vpc_id                  = module.networking.vpc_id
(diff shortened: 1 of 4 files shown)


terraform/production/main.tf
```diff
@@ -127,7 +127,7 @@ module "redis" {
 
   prefix                  = local.prefix
   ecs_security_group_id   = module.application.ecs_security_group_id
-  node_type               = "cache.t2.micro"
+  node_type               = "cache.t4g.micro"
```
SamSeed-Softwire, line 130: Just updating prod and dev now as well given they're currently set to the same `node_type` as staging. This can be reviewed later of course.

terraform/development/main.tf
```diff
@@ -127,7 +127,7 @@ module "redis" {
 
   prefix                  = local.prefix
   ecs_security_group_id   = module.application.ecs_security_group_id
-  node_type               = "cache.t2.micro"
+  node_type               = "cache.t4g.micro"
```
SamSeed-Softwire, line 130: Just updating prod and dev now as well given they're currently set to the same `node_type` as staging. This can be reviewed later of course.

RachaelBooth: Sounds sensible - I think we explicitly have a ticket for reviewing sizes for production, and assume that we'll cover dev when we look at review apps.

RachaelBooth (commented):
Looks good to me.

I think with our current setup you can restore the ignore changes before merging, as long as the terraform has been applied at least once without - but when you do that apply you'll need to make sure the image has been updated to what we're actually running.

RachaelBooth (approved):

PR #39, opened 2023-09-14
CLDC-2777: Set up SNS

SamSeed-Softwire:
Will do dev and prod once the rest of this PR is finalised.

### DIFF
--- a/terraform/modules/monitoring/policy.tf
+++ b/terraform/modules/monitoring/policy.tf
@@ -17,6 +17,14 @@ data "aws_iam_policy_document" "this" {
     ]
 
     principals {
+      type        = "AWS"
+      identifiers = ["*"]
+    }
+
+    condition {
+      test     = "ArnLike"
+      variable = "aws:SourceArn"
+      values   = []
     }
   }
 }
--- a/terraform/modules/monitoring/policy.tf
+++ b/terraform/modules/monitoring/policy.tf
@@ -17,14 +17,16 @@ data "aws_iam_policy_document" "this" {
     ]
 
     principals {
-      type        = "AWS"
-      identifiers = ["*"]
+      type        = "Service"
+      identifiers = ["cloudwatch.amazonaws.com"]
     }
 
     condition {
-      test     = "ArnLike"
-      variable = "aws:SourceArn"
-      values   = []
+      test     = "StringEquals"
+      variable = "aws:SourceAccount"
+      values   = [data.aws_caller_identity.current.account_id]
     }
   }
 }
+
+data "aws_caller_identity" "current" {}
--- a/terraform/modules/monitoring/policy.tf
+++ b/terraform/modules/monitoring/policy.tf
@@ -1,0 +1,22 @@
+resource "aws_sns_topic_policy" "this" {
+  arn = aws_sns_topic.this.arn
+
+  policy = data.aws_iam_policy_document.this.json
+}
+
+data "aws_iam_policy_document" "this" {
+  statement {
+    actions = [
+      "SNS:Publish"
+    ]
+
+    effect = "Allow"
+
+    resources = [
+      aws_sns_topic.this.arn
+    ]
+
+    principals {
+    }
+  }
+}
--- a/terraform/modules/monitoring/topic.tf
+++ b/terraform/modules/monitoring/topic.tf
@@ -1,0 +1,3 @@
+resource "aws_sns_topic" "this" {
+  name = var.prefix
+}
--- a/terraform/modules/monitoring/topic.tf
+++ b/terraform/modules/monitoring/topic.tf
@@ -1,3 +1,4 @@
+#tfsec:ignore:aws-sns-enable-topic-encryption:encryption at rest is overkill for the type of data we will be sending to the topic
 resource "aws_sns_topic" "this" {
   #checkov:skip=CKV_AWS_26:encryption at rest is overkill for the type of data we will be sending to the topic
   name = var.prefix
--- a/terraform/modules/monitoring/topic.tf
+++ b/terraform/modules/monitoring/topic.tf
@@ -1,3 +1,4 @@
 resource "aws_sns_topic" "this" {
+  #checkov:skip=CKV_AWS_26:encryption at rest is overkill for the type of data we will be sending to the topic
   name = var.prefix
 }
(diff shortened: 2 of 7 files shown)


terraform/modules/monitoring/policy.tf
```diff
@@ -0,0 +1,30 @@
+resource "aws_sns_topic_policy" "this" {
```
SamSeed-Softwire, line 1: I could have also just used regular IAM policies here. See [here](https://docs.aws.amazon.com/sns/latest/dg/sns-using-identity-based-policies.html#iam-and-sns-policies) for a comparison of the approaches. In retrospect I think it might actually have been better to use regular IAM policies (rather than `aws_sns_topic_policy`) - like how it was done in EPB. The way I've done it here does allow for cross-account SNS publishing, but I don't think that will be needed.

I wanted to get a second opinion though about whether the other way was indeed better, and even if it was whether it was worth the time to change it (probably yes IMO).

terraform/modules/monitoring/policy.tf
```diff
@@ -0,0 +1,30 @@
...
+
+    principals {
+      type        = "AWS"
+      identifiers = ["*"]
+    }
+
+    condition {
+      test     = "ArnLike"
+      variable = "aws:SourceArn"
+      values   = []
```
SamSeed-Softwire, line 27: Currently nothing is allowed to publish - my plan was to add ARNs here for resources as they are created.

terraform/modules/monitoring/topic.tf
```diff
@@ -0,0 +1,5 @@
+#tfsec:ignore:aws-sns-enable-topic-encryption:encryption at rest is overkill for the type of data we will be sending to the topic
+resource "aws_sns_topic" "this" {
```
SamSeed-Softwire, line 2: AFAIK there will only be one SNS topic which we use for everything. Happy to use a disambiguating name now though if that's not (or might not) be the case.

terraform/modules/monitoring/policy.tf
RachaelBooth: Overall I don't think it matters - the main concern is probably how easy it is to add new things that can publish to the topic (particularly for e.g. review apps where we might end up with one topic and several instances of other things publishing notifications)

terraform/staging/main.tf
```diff
@@ -122,6 +122,12 @@ module "networking" {
   vpc_flow_cloudwatch_log_expiration_days = 90
 }
 
+module "monitoring" {
```
Chirag-Bhatti, line 125: Believe we might need to add this in dev and prod as well? 😊

SamSeed-Softwire: Ah, sorry, knew I'd forgotten to write something on the PR - basically I'm generally just doing stuff for staging, then once the PR has been approved and I know exactly what staging should look like only then do I replicate for dev and prod. Helps to reduce the amount of work I have to do!

terraform/modules/monitoring/topic.tf
Chirag-Bhatti: A single topic feels right to me (we're only using this for alarms / notifications to slack)

terraform/modules/monitoring/policy.tf
Chirag-Bhatti: Looks good (presuming it's the arn's of the `aws_cloudwatch_metric_alarm`'s that will be added here)

Chirag-Bhatti:
@RachaelBooth / @SamSeed-Softwire, it might be out of scope for ticket/MVP, but I wanted to check if we should consider covering the case where the alarm can't be delivered to the subscriber from SNS? Should we setup a dead letter queue for example? Some info [here](https://docs.aws.amazon.com/sns/latest/dg/sns-dead-letter-queues.html)

I suppose the alarm not reaching Slack might not be so bad if we also send out email notifications for example. Also, if the alarm notification reaches much later because the issue preventing delivery was temporary, it might not be as useful anymore too. But we also may want to ensure an important alarm isn't missed because there was only one medium of knowing (outside of checking the AWS console)

SamSeed-Softwire:
> @RachaelBooth / @SamSeed-Softwire, it might be out of scope for ticket/MVP, but I wanted to check if we should consider covering the case where the alarm can't be delivered to the subscriber from SNS? Should we setup a dead letter queue for example? Some info [here](https://docs.aws.amazon.com/sns/latest/dg/sns-dead-letter-queues.html)
> 
> I suppose the alarm not reaching Slack might not be so bad if we also send out email notifications for example. Also, if the alarm notification reaches much later because the issue preventing delivery was temporary, it might not be as useful anymore. But we may want to ensure an important alarm isn't missed, because there was only one medium of knowing (outside of checking the AWS console)
> 
> What's your thoughts?

@Chirag-Bhatti I think that sounds sensible. From reading about it it seems that DLQs are at the subscription-level, so even if we do want them they are out of scope for this ticket (which you might have already realised). If we do want DLQs then I guess we should do them as part of each ticket where we're creating a subscription.

terraform/modules/monitoring/policy.tf
RachaelBooth: Makes sense to me

RachaelBooth (approved):

RachaelBooth:
> > @RachaelBooth / @SamSeed-Softwire, it might be out of scope for ticket/MVP, but I wanted to check if we should consider covering the case where the alarm can't be delivered to the subscriber from SNS? Should we setup a dead letter queue for example? Some info [here](https://docs.aws.amazon.com/sns/latest/dg/sns-dead-letter-queues.html)
> > I suppose the alarm not reaching Slack might not be so bad if we also send out email notifications for example. Also, if the alarm notification reaches much later because the issue preventing delivery was temporary, it might not be as useful anymore. But we may want to ensure an important alarm isn't missed, because there was only one medium of knowing (outside of checking the AWS console)
> > What's your thoughts?
> 
> @Chirag-Bhatti I think that sounds sensible. From reading about it it seems that DLQs are at the subscription-level, so even if we do want them they are out of scope for this ticket (which you might have already realised). If we do want DLQs then I guess we should do them as part of each ticket where we're creating a subscription.

The interesting thing about dead letter queues with this kind of use case is that our only subscription is going to be "notify us of alarms" and the way we'd find out that something had hit the dead letter queue would be a similar alarm and thus likely to be subject to the same problem. (Assuming it's not a message specific problem, but I wouldn't expect it to be for this type of message)

We ought to consider one at some point, perhaps with a different notification mechanism? I think it's relatively low down the priority list though.

PR #46, opened 2023-09-20
Fix & improve load balancer + cloudfront

SamSeed-Softwire:
Final edits necessary to get all the front door stuff working + more secure:
- Use http, not https, between load balancer and application
- Remove obsolete checkov and tfsec ignores
- Restrict the load balancer ingress to only accept https traffic from cloudfront

Will `fmt` once everything else has been finalised.


RachaelBooth (approved):
Looks good once format passes.

PR #47, opened 2023-09-20
CLDC-2787: Basic RDS alarms

SamSeed-Softwire:
https://dluhcdigital.atlassian.net/browse/CLDC-2787


terraform/modules/monitoring/alarms.tf
```diff
@@ -77,3 +77,65 @@ resource "aws_cloudwatch_metric_alarm" "sidekiq_memory" {
...
+  }
+
+  metric_query {
+    id = "freeStorageSpace"
+
+    metric {
+      metric_name = "FreeStorageSpace"
+      namespace   = "AWS/RDS"
+      period      = 300
+      stat        = "Minimum"
```
SamSeed-Softwire, line 124: I'm assuming storage space isn't really volatile over very short timescales, so I went for minimum here as the stat. If for some reason we're expecting the storage space to quickly fluctuate as part of how the db functions, we should probably change this to average or something.

terraform/staging/main.tf
```diff
@@ -147,6 +148,8 @@ module "monitoring" {
 
   prefix               = local.prefix
   app_service_name     = module.application.app_service_name
+  database_allocated_storage = local.database_allocated_storage
+  database_id = module.database.rds_id
```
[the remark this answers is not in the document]
RachaelBooth: I think the aim (that we don't quite actually fulfil) is that we use `database` when we don't care what type of database it is (e.g. the app doesn't care whether this is an rds instance or aurora or something completely different) and `rds` when it matters that it's specifically that.

terraform/staging/outputs.tf
```diff
@@ -1,9 +0,0 @@
-output "cloudfront_certificate_validation" {
```
RachaelBooth, line 1: This looks like an accident?

terraform/modules/monitoring/alarms.tf
```diff
@@ -77,3 +77,55 @@ resource "aws_cloudwatch_metric_alarm" "sidekiq_memory" {
     ServiceName = var.sidekiq_service_name
   }
 }
+
```
RachaelBooth, line 80: I have some thoughts on where these live prompted by thinking through review apps;

Notably, we're going to need to separate things that are per review app from things that will be shared between review apps. Since the services will be per app and the database instance will not, I think this forces us not to have these rds alarms in the same module as the ecs ones.

Which gives us two approaches - either we have two monitoring modules, one covering shared things and one covering per app things, or (I think my preferred option) we move the alarms to live alongside the things they are monitoring. Either way the sns queue / notifications setup lives in a monitoring module.

We may as well do this now rather than leaving it to be review apps work.

terraform/modules/monitoring/alarms.tf
```diff
@@ -77,3 +77,65 @@ resource "aws_cloudwatch_metric_alarm" "sidekiq_memory" {
...
+  alarm_name                = "${var.prefix}-rds-storage"
+  comparison_operator       = "LessThanOrEqualToThreshold"
+  datapoints_to_alarm       = 1
+  evaluation_periods        = 1
+  ok_actions                = [aws_sns_topic.this.arn]
+  threshold                 = 15
+  insufficient_data_actions = []
+
+  metric_query {
+    id = "freeStorageSpacePercentage"
```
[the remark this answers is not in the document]
RachaelBooth: Rather than needing the two metric queries, could this just compare the free storage space with the database allocated storage * 15/100 ?

terraform/staging/outputs.tf
SamSeed-Softwire: @RachaelBooth sorry, forgot to write a comment on here. Chirag suggested we might not need these outputs any more (at least for staging) since the infra exists now. So I thought it would be fine to remove since we're seeing those outputs whenever we run apply. Happy to add back in if you disagree!

RachaelBooth: I think we might want them for disaster recovery purposes (i.e. when building an env from scratch we do want them, although that would also be complicated by the ordering problem on validating the certificates). I would leave them in and consider at the point of writing up disaster recovery stuff / how to build a new env / doing cleanup at the end of the project.

terraform/staging/main.tf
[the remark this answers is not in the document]
SamSeed-Softwire: Thanks @RachaelBooth. I've raised a 'could have' ticket to standardise this: https://dluhcdigital.atlassian.net/browse/CLDC-2819.

PR #49, opened 2023-09-25
CLDC-2618: Infra for migrating DB from PaaS to AWS

Chirag-Bhatti:
- Separate ECR in meta account for the DB migration image

- Separate ECS cluster, service and tasks
- A Dockerfile to run on ECS and store in the ECR
- A script for running the DB migration process
- One task that runs continuously but doesn't do anything, so we can exec into it
- Another task that runs the migration script (started on an ad-hoc basic in the AWS console)
- Basic ECS logging (to allow debugging)

- Secrets for access to Gov PaaS
- Enable ECS task execution role to access these secrets
- Security groups to allow ECS to connect to the DB, connect to Gov PaaS (including use of the conduit plugin) and get secrets

- A flag at the environment entrypoint in the terraform so we can opt to create / not create this infra


terraform/modules/rds_migration/security_groups.tf
```diff
@@ -0,0 +1,27 @@
...
+resource "aws_vpc_security_group_ingress_rule" "db_migration_ecs_ingress" {
+  description                  = "Allow ingress on port ${var.database_port} of the db from the db migration ecs"
+  ip_protocol                  = "tcp"
+  from_port                    = var.database_port
+  to_port                      = var.database_port
+  referenced_security_group_id = aws_security_group.ecs.id
+  security_group_id            = var.db_security_group_id
+}
+
+resource "aws_vpc_security_group_egress_rule" "general_egress" {
```
Chirag-Bhatti, line 20: The `cf conduit` plugin seems to consistently use port 7080 on my machine in order to connect with the DB service when going to make a `pg dump` file. However on ECS it seems to be using a different port and I haven't been able to work out which and if it consistently uses that. I've tried a rule with all ports 7080 and above based on the conduit logic [here](https://github.com/alphagov/paas-cf-conduit/blob/b16106aa5089599280711403773f130b9493aeb0/main.go#L82), but that didn't work either! It didn't seem worth the time to try and pin point.

As a result, I've had to allow egress to all ports in order to make conduit work 😥 (this also allows ECS to connect to the DB and get secrets, otherwise there would be separate rules for these). I think this is ok given we'll run this on an ad-hoc basis a handful of times, and destroy the infra when we're not migrating from PaaS

db-migration/run.sh
```diff
@@ -0,0 +1,35 @@
...
+
+if cf conduit ${CF_SERVICE} -- pg_dump --file dumpfile --no-acl --encoding utf8 --clean --no-owner --if-exists -Fc; then
+  echo 'Dump file of PaaS DB created successfully'
+else
+  echo 'ERROR trying to conduit into the PaaS DB or creating the dump file'
+  exit 1
+fi
+
+while [ ! -e  "$dumpfile"]; do
+  echo "Waiting for dump file to be ready for pg restore"
```
Chirag-Bhatti, line 22: I noticed on my local machine that sometimes the dump file is still increasing in size on the drive after the command has finished, so I'm hoping this will allow the file to be fully created before proceeding to use it for the restore

terraform/modules/rds_migration/role_task_execution.tf
```diff
@@ -0,0 +1,26 @@
...
+          aws_secretsmanager_secret.cloudfoundry_password.arn,
+          aws_secretsmanager_secret.cloudfoundry_space.arn,
+          aws_secretsmanager_secret.cloudfoundry_service.arn
+        ]
+      }
+    ]
+  })
+}
+
+resource "aws_iam_role_policy_attachment" "secret_access" {
```
Chirag-Bhatti, line 23: This effectively concatenates the above policy to the policies already existing on task execution role from the `application` module

terraform/modules/rds_migration/logging.tf
```diff
@@ -0,0 +1,12 @@
+# TODO CLDC-2542 remove/modify this temporary logging setup
+#tfsec:ignore:aws-cloudwatch-log-group-customer-key:TODO CLDC-2542 create and encrypt with a KMS key if retaining this log group
+resource "aws_cloudwatch_log_group" "this" {
```
Chirag-Bhatti, line 3: FYI @SamSeed-Softwire - in case it impacts anything on [CLDC-2779](https://dluhcdigital.atlassian.net/browse/CLDC-2779) (improve cloudwatch log groups)

terraform/modules/ecr/main.tf
```diff
@@ -20,6 +20,17 @@ resource "aws_ecr_repository" "this" {
   }
 }
 
+#tfsec:ignore:aws-ecr-repository-customer-key:encryption using KMS CMK not required
+resource "aws_ecr_repository" "db_migration" {
+  #checkov:skip=CKV_AWS_136:encryption using KMS not required
+  name                 = "db-migration"
+  image_tag_mutability = "IMMUTABLE"
```
Chirag-Bhatti, line 27: I think it maybe preferable in this case to not have IMMUTABLE images and let them have the same "latest" tag? Any thoughts on this? 

I'm also wondering if we should update our pipelines to automatically build the DB migration image and push it to the ECR, but I feel this is probably unnecessary given we're expecting to only use it a handful of times max.

db-migration/run.sh
SamSeed-Softwire: Doesn't the `-e` expression just check for file existence? Surely that won't help with checking its size has stabilised?

SamSeed-Softwire (commented):
I still have more to review, just thought it would be better to add individual comments rather than saving them all up.

db-migration/run.sh
```diff
@@ -0,0 +1,35 @@
...
+  sleep 1
+done
+
+echo 'Dump file ready for pg restore'
+echo 'Starting pg restore of AWS DB'
+
+if pg_restore -v -d ${DATABASE_URL} -j 4 --no-acl --clean --no-owner --if-exists -Fc dumpfile; then
```
SamSeed-Softwire, line 29: Similarly to with `pg_dump` do we want to consider using the `--create` option?

terraform/modules/application/outputs.tf
```diff
@@ -2,3 +2,23 @@ output "ecs_security_group_id" {
...
+  value       = aws_iam_role.task_execution.name
+  description = "The id of the ecs task execution role"
+}
+
+output "ecs_task_role_arn" {
+  value       = aws_iam_role.task.arn
+  description = "The arn of the ecs task role"
+}
+
+output "sidekiq_service_name" {
```
SamSeed-Softwire, line 21: Is this getting used anywhere?

terraform/modules/ecr/main.tf
```diff
@@ -10,7 +10,7 @@ terraform {
 }
 
 #tfsec:ignore:aws-ecr-repository-customer-key:encryption using KMS CMK not required
-resource "aws_ecr_repository" "this" {
+resource "aws_ecr_repository" "core" {
```
SamSeed-Softwire, line 13: I would maybe have called this "app" but no biggie.

terraform/modules/ecr/main.tf
SamSeed-Softwire: Why not have immutable images? For convenience? If we do have immutable images, we'll need to think of a sensible tagging strategy (I'm assuming we won't want commit refs, as the contents of the image will be determined by the contents of the db rather than the state of the code at the time of running the Dockerfile). But we could definitely use timestamps to easily generate unique names.

terraform/modules/rds_migration/ecs.tf
```diff
@@ -0,0 +1,128 @@
+#tfsec:ignore:aws-ecs-enable-container-insight:TODO CLDC-2542 enable container insights if necessary for logging/monitoring
+resource "aws_ecs_cluster" "this" {
+  #checkov:skip=CKV_AWS_65:TODO CLDC-2542 enable container insights if necessary for logging/monitoring
+  name = "${var.prefix}-db-migration"
```
SamSeed-Softwire, line 4: Do we want to rename the main ecs cluster to something like `${var.prefix}-app`?

terraform/modules/rds_migration/ecs.tf
```diff
@@ -0,0 +1,128 @@
...
+    size_in_gib = var.ecs_task_ephemeral_storage
+  }
+
+  runtime_platform {
+    operating_system_family = "LINUX"
+    cpu_architecture        = "X86_64"
+  }
+
+  lifecycle {
+    # The image will be updated by deployments - irritatingly we can't ignore changes just to the image
```
SamSeed-Softwire, line 52: Will the image be updated by deployments? That's tbc right.

terraform/modules/rds_migration/ecs.tf
```diff
@@ -0,0 +1,128 @@
...
+  memory                   = var.db_migration_task_memory #MiB
+  network_mode             = "awsvpc"
+  requires_compatibilities = ["FARGATE"]
+  task_role_arn            = var.ecs_task_role_arn
+
+  container_definitions = jsonencode([
+    {
+      name      = "db-migration"
+      essential = true
+      image     = var.ecr_repository_url
```
SamSeed-Softwire, line 21: Currently this won't return a valid image AFAIK. For the regular ECS app task this gets updated by the deployment pipeline I believe.

terraform/modules/rds_migration/outputs.tf
```diff
@@ -0,0 +1,4 @@
+output "db_migration_ecs_security_group_id" {
```
SamSeed-Softwire, line 1: Is this being used anywhere?

terraform/modules/ecr/main.tf
Chirag-Bhatti: Yeah more for convenience. At the moment we just tag the image with `latest` and so anytime we want to put another migration image in the ECR, we have to do delete the old `latest` first due to immutability. And in this case I'm not sure we need multiple images with different tags. We're also building and uploading the images locally, so any tagging strategy needs to be easy to do here.

db-migration/run.sh
Chirag-Bhatti: It does! My thinking was that this would be sufficient that by the time we've seen it exists it will probably be ready pretty quickly after that for use. It's a non-trivial problem to work out when the file is finished being written to however 😢, one option I know of is `lsof` (but not tried this), keeping checking the size of the file, or just sleep for an arbitrary amount of time to give it breathing space 😁. We can try one of these or just see how we get on, not seen any issues with the above yet but it's more likely to occur when the DBs are big!

SamSeed-Softwire: Fair enough, worth discussing with Rachael I think if you haven't already.

db-migration/run.sh
```diff
@@ -0,0 +1,35 @@
...
+  echo 'Ran pg restore on AWS DB successfully'
+  exit 0
+else
+  echo 'ERROR running pg restore on AWS DB. Check if the errors reported are important or harmless as the restore may still have completed ok'
+  exit 1
+fi
```
RachaelBooth, line 35: Question: if an error does happen during this process, what can we do to deal with it? (It may be that our best option is to start again if anything important has failed, but we should make sure we know that)

db-migration/run.sh
```diff
@@ -0,0 +1,35 @@
...
+if cf login -a api.london.cloud.service.gov.uk -u ${CF_USERNAME} -p ${CF_PASSWORD} -s ${CF_SPACE}; then
+  echo 'Logged into Gov PaaS successfully'
+else
+  echo 'ERROR could not log into Gov Paas'
+  exit 1
+fi
+
+echo "Setting up conduit into Gov PaaS DB and making dump file"
+
+if cf conduit ${CF_SERVICE} -- pg_dump --file dumpfile --no-acl --encoding utf8 --clean --no-owner --if-exists -Fc; then
```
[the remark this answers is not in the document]
Chirag-Bhatti: Ah yes checked through which ones I thought we'd likely need based off docs and EPB. See notes below:

`--compress` - not required as we're using `-Fc`, which is the custom-format and compressed by default
`--create` - used with `--clean` will drop the DB before recreating it. This caused weird errors when I was testing with staging. I think it's safer to make a green DB using a rake first but I also didn't spend too long investigating
`--jobs` - we could set this but as we'll connect to the prod DB we don't want to occupy its performance capacity or connections too much (unless noone is using it at the time). I'll check if I can get the vCPUs of the PaaS DB (can't recall if we had this info but it's generally advisable for these to match) and see if we could set this
`--lock-wait-timeout` - this didn't seem necessary but I'm unsure when we'd need it either. Probably best to leave and investigate if we find any issue
`--verbose` - good spot. I enabled this on `pg_restore` and missed the same here, I think it'd be useful

db-migration/run.sh
Chirag-Bhatti: See comment above on `pg_dump` (same thing applies here) 👍

terraform/modules/application/outputs.tf
Chirag-Bhatti: Nope looks like an accidental copy, will remove!

terraform/modules/ecr/main.tf
Chirag-Bhatti: Not sure why my comment appeared after Rachael's when I wrote it a while before 😄, nonetheless I'll go with mutable and remove the ignore changes in the terraform 👍

terraform/modules/rds_migration/ecs.tf
Chirag-Bhatti: Rachael has confirmed we'll do this manually, so I've removed all these lifecycle blocks and comments about images / deployments 👍

terraform/modules/rds_migration/ecs.tf
```diff
@@ -0,0 +1,128 @@
...
+    # to force the update, ensuring the referenced image is the correct current one
+    ignore_changes = [container_definitions]
+  }
+}
+
+# This ECS task just runs continuously in the background doing nothing. This is so we can exec into it later and check the DB
+resource "aws_ecs_task_definition" "exec_placeholder" {
+  #checkov:skip=CKV_AWS_336:using readonlyRootFilesystem to true breaks the app, as it needs to write to app/tmp/pids for example
+  family                   = "${var.prefix}-exec-placeholder"
+  cpu                      = var.db_migration_task_cpu
```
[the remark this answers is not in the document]
Chirag-Bhatti: Technically it's unlikely, however I wanted to leave the possibility open of doing any intensive operations using this task if we need to debug something. As this infra should live temporarily I thought that's probably OK, and also means we don't have to remember to tweak if it comes to it 😀

terraform/modules/rds_migration/outputs.tf
Chirag-Bhatti: Nice spot! It isn't, it's a leftover from some implementation I was trying. Will remove.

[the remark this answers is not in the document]
Chirag-Bhatti: Nope, will be removed!

terraform/modules/ecr/main.tf
Chirag-Bhatti: Ah yes, in this case the `name` property of the resource was `core` so I thought it would be clearer if I matched that 😀

SamSeed-Softwire: I probably would have given the resource the name `app` as well! But again, just personal preference and not worth deleting the ECR over.

terraform/modules/ecr/main.tf
```diff
@@ -20,6 +20,17 @@ resource "aws_ecr_repository" "this" {
   }
 }
 
+#tfsec:ignore:aws-ecr-repository-customer-key:encryption using KMS CMK not required
+resource "aws_ecr_repository" "db_migration" {
```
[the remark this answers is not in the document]
Chirag-Bhatti: Yeah that's a good point! Originally I was thinking we may want to persist this but on reflection it's better not to. I'll refactor so this ECR is only made if we set a flag to (similarly to the `rds_migration` module)

terraform/modules/rds_migration/logging.tf
```diff
@@ -0,0 +1,12 @@
+# TODO CLDC-2542 remove/modify this temporary logging setup
```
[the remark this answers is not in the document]
Chirag-Bhatti: Nice spot! I'll update these. FYI @SamSeed-Softwire I'll be putting TODOs here against [CLDC-2779](https://dluhcdigital.atlassian.net/browse/CLDC-2779)

db-migration/run.sh
Chirag-Bhatti: Good question! My thinking here is that all the steps are pretty critical so if one fails, we'll probably want to investigate and start over. I think it is possible to run the` pg_restore` over a DB (believe it is idempotent although it might report errors about things already existing), however depending on the kind of error we may want to clean out the DB and start from green again.

Also we have the exec placeholder task in order to be able to re-run the commands and try and diagnose in more detail what's going on.

terraform/modules/rds_migration/ecs.tf
```diff
@@ -0,0 +1,128 @@
...
+
+  lifecycle {
+    # The image will be updated by deployments - irritatingly we can't ignore changes just to the image
+    # If changing other aspects of the container definition we'll need to temporarily not ignore changes
+    # to force the update, ensuring the referenced image is the correct current one
+    ignore_changes = [container_definitions]
+  }
+}
+
+resource "aws_ecs_service" "db_migration" {
```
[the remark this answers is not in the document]
Chirag-Bhatti: If I understand your comment correctly, the service is running the `exec_placeholder` task, which I want to keep running (it doesn't do anything) whenever the db migration infra has been created, so that we can exec into it anytime to debug the migration process / check things on the db

Maybe the naming of the service is misleading? Lmk if I've misunderstood and we shout chat to discuss 😃

terraform/modules/rds_migration/logging.tf
```diff
@@ -0,0 +1,12 @@
+# TODO CLDC-2542 remove/modify this temporary logging setup
+#tfsec:ignore:aws-cloudwatch-log-group-customer-key:TODO CLDC-2542 create and encrypt with a KMS key if retaining this log group
+resource "aws_cloudwatch_log_group" "this" {
+  #checkov:skip=CKV_AWS_158:TODO CLDC-2542 create and encrypt with a KMS key if retaining this log group
+  #checkov:skip=CKV_AWS_338:TODO CLDC-2542 retaining logs for at least 1year greater than necessary for debugging ecs
+  name              = "${var.prefix}-db-migration-ecs-debugging-log"
```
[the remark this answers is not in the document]
Chirag-Bhatti: Ah so the application ECS has a similar name currently (`"${var.prefix}-ecs-debugging-log"`) but maybe this doesn't need "debugging" in the name anymore, since we intend to use Cloudwatch as our logging solution (we originally put this in just to help us debug the ECS)

So I think the app ECS log group could be called: "${var.prefix}-app-ecs"
And the db miration ECS one be called: "${var.prefix}-db-migration-ecs"

Do you agree @SamSeed-Softwire? If so lmk and I can update them here if you prefer 😀

db-migration/run.sh
Chirag-Bhatti: Documentation updates done 👍

terraform/modules/rds_migration/ecs.tf
```diff
@@ -0,0 +1,128 @@
...
+          awslogs-region        = "eu-west-2"
+          awslogs-stream-prefix = var.prefix
+        }
+      }
+
+      secrets = [
+        { Name = "DATABASE_URL", valueFrom = var.database_connection_string_arn },
+        { Name = "CF_PASSWORD", valueFrom = aws_secretsmanager_secret.cloudfoundry_password.arn },
+        { Name = "CF_SERVICE", valueFrom = aws_secretsmanager_secret.cloudfoundry_service.arn },
+        { Name = "CF_SPACE", valueFrom = aws_secretsmanager_secret.cloudfoundry_space.arn },
```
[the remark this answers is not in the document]
Chirag-Bhatti: I felt like these were potentially also sensitive (I was obfuscating all the information needed to interact with the PaaS Production DB) but indeed not as much as username and password. If not a secret, would you be ok with hardcoding these into terraform or were you thinking something else?

terraform/modules/rds_migration/logging.tf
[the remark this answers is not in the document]
SamSeed-Softwire: Yep sounds good to me!

terraform/modules/rds_migration/ecs.tf
```diff
@@ -0,0 +1,128 @@
+#tfsec:ignore:aws-ecs-enable-container-insight:TODO CLDC-2542 enable container insights if necessary for logging/monitoring
+resource "aws_ecs_cluster" "this" {
+  #checkov:skip=CKV_AWS_65:TODO CLDC-2542 enable container insights if necessary for logging/monitoring
+  name = "${var.prefix}-db-migration"
+}
+
+resource "aws_ecs_task_definition" "ad_hoc_db_migration" {
+  #checkov:skip=CKV_AWS_336:using readonlyRootFilesystem to true breaks the app, as it needs to write to app/tmp/pids for example
```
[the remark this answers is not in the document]
Chirag-Bhatti: Unfortunately this does break the task (e.g. `cf login` needs to write to a root folder which affects the `db_migration` task, and it maybe something we try to do when `exec`'d into the `exec_placeholder` task, so this needs to be set to false for both)

I'll add checkov ignore comments about this 👍

db-migration/run.sh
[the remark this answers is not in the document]
Chirag-Bhatti: I've updated the `jobs` and `verbose `options on the `pg_dump` command 👍

db-migration/run.sh
Chirag-Bhatti: Ok I think I've made an improvement to this, by first checking the file exists (thought it might be useful to keep this anyway) and then running a `fuser` command on a loop to check if the dumpfile is being accessed by any processes. Testing hasn't shown if this definitely useful as the dumpfile is so quick to write it doesn't end up reporting anything, but we can see how we get on when we do a staging migration

terraform/modules/rds_migration/ecs.tf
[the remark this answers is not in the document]
RachaelBooth: The naming of the service is indeed misleading, since it's not actually running the db migration.

But what I was trying to get at was that I don't think the exec_placeholder task needs to be left running for the entire time that we have the db migration infrastructure in existence (particularly as this will be longer than the time that the actual db migration is running). Unless I've missed something, we should only care about it when we're actually interacting with it right? So I was thinking that we'd spin it up and down as a task when we needed.

terraform/modules/rds_migration/ecs.tf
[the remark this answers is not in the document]
RachaelBooth: Possibly hardcoding in terraform, possibly as vars (although that's always a bit annoying to deal with), and passed into the container as env variables rather than secrets. As I say it's not really a problem to do it this way, just adds a manual step of setting them (and checking if they're correct if we need to).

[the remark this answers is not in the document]
Chirag-Bhatti: I've set them up as vars and updated the documentation to say these should be checked and updated if necessary 👍 (for staging and prod these should be correct already, and for dev we don't expect to use it to for this but I have given a placeholding value for the cf service anyway)

terraform/modules/rds_migration/ecs.tf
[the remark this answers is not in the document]
Chirag-Bhatti: Ah yes I see what you mean! My thinking was that the infra as a whole shouldn't exist much longer than it needs to, but when it does we should have an exec task ready for any quick debugging/checking we need to do. Nonetheless it's not so hard to spin up so I've updated the ECS to desire 0 `db_migration` tasks (making the naming less confusing) and updated the documentation so we can launch the `exec_placeholder` task as and when required.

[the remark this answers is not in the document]
RachaelBooth: Can we just skip the service in that case? We can spin up the task on it's own with a cli command (with the `--enable-execute-command` option) rather than needing to change the number on a service.

I'm sort of expecting that infra does end up existing for a while, because we don't want to delay the start of actually migrating if something is going wrong with creating it so I think setting it up is an ahead-of-time thing. (And we probably don't bother to destroy it after creating it for a test)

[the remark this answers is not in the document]
Chirag-Bhatti: I've removed the ecs service from the implementation, and updated the process documentation so that we can launch the tasks as required using the AWS CLI (without ever needing a service) 😀👍

terraform/meta/main.tf
```diff
@@ -69,6 +70,15 @@ module "ecr" {
   allow_access_by_roles = ["arn:aws:iam::107155005276:role/core-staging-task-execution"]
 }
 
+module "ecr_rds_migration" {
+  source = "../modules/ecr_rds_migration"
+
+  count = local.create_db_migration_infra ? 1 : 0
+
+  # This will need updating to include dev and production roles
+  allow_access_by_roles = ["arn:aws:iam::107155005276:role/core-staging-task-execution"]
```
SamSeed-Softwire, line 79: Maybe we could make this a local var now that it's used twice in the file?

SamSeed-Softwire (commented):
Lgtm except for the checkov warning.

RachaelBooth (approved):
LGTM, should just need a checkov skip (since we will use that security group for running tasks, it's just not attached to a resource here)

PR #50, opened 2023-09-26
CLDC-2785: App running tasks alarm

SamSeed-Softwire:
As discussed, I've implemented:
- An alarm which looks for the "Essential container in task exited" event a certain number of times over a certain period
- An alarm which looks for any ECS Service Action issues (WARN or ERROR events)
- An alarm which checks for unhealthy hosts, and an alarm which checks whether the right number of healthy hosts

### DIFF
--- a/terraform/modules/application/alarms.tf
+++ b/terraform/modules/application/alarms.tf
@@ -47,7 +47,7 @@ resource "aws_cloudwatch_metric_alarm" "app_tasks_exited" {
   metric_name               = "TriggeredRules"
   namespace                 = "AWS/Events"
   ok_actions                = [var.sns_topic_arn]
-  period                    = 8 * 60
+  period                    = 8 * 60 # By causing the app to fail repeatedly (with both 2 tasks and 4 tasks running) I found that double the desired count would reliably fail in an 8 minute period.
   statistic                 = "Sum"
   threshold                 = 2 * var.app_task_desired_count
   insufficient_data_actions = []
--- a/terraform/modules/application/alarms.tf
+++ b/terraform/modules/application/alarms.tf
@@ -38,6 +38,25 @@ resource "aws_cloudwatch_metric_alarm" "app_memory" {
   }
 }
 
+resource "aws_cloudwatch_metric_alarm" "app_tasks_exited" {
+  alarm_actions             = [var.sns_topic_arn]
+  alarm_name                = "${var.prefix}-app-tasks-exited"
+  comparison_operator       = "GreaterThanOrEqualToThreshold"
+  datapoints_to_alarm       = 2 * var.app_task_desired_count
+  evaluation_periods        = 8
+  metric_name               = "TriggeredRules"
+  namespace                 = "AWS/Events"
+  ok_actions                = [var.sns_topic_arn]
+  period                    = 60
+  statistic                 = "Sum"
+  threshold                 = 1
+  insufficient_data_actions = []
+
+  dimensions = {
+    RuleName = aws_cloudwatch_event_rule.app_task_exited.name
+  }
+}
+
 resource "aws_cloudwatch_metric_alarm" "sidekiq_cpu" {
   alarm_actions             = [var.sns_topic_arn]
   alarm_name                = "${var.prefix}-sidekiq-cpu"
--- a/terraform/modules/application/alarms.tf
+++ b/terraform/modules/application/alarms.tf
@@ -42,14 +42,14 @@ resource "aws_cloudwatch_metric_alarm" "app_tasks_exited" {
   alarm_actions             = [var.sns_topic_arn]
   alarm_name                = "${var.prefix}-app-tasks-exited"
   comparison_operator       = "GreaterThanOrEqualToThreshold"
-  datapoints_to_alarm       = 2 * var.app_task_desired_count
-  evaluation_periods        = 8
+  datapoints_to_alarm       = 1
+  evaluation_periods        = 1
   metric_name               = "TriggeredRules"
   namespace                 = "AWS/Events"
   ok_actions                = [var.sns_topic_arn]
-  period                    = 60
+  period                    = 8 * 60
   statistic                 = "Sum"
-  threshold                 = 1
+  threshold                 = 2 * var.app_task_desired_count
   insufficient_data_actions = []
 
   dimensions = {
--- a/terraform/modules/application/alarms.tf
+++ b/terraform/modules/application/alarms.tf
@@ -57,6 +57,25 @@ resource "aws_cloudwatch_metric_alarm" "app_tasks_exited" {
   }
 }
 
+resource "aws_cloudwatch_metric_alarm" "app_service_action_problem" {
+  alarm_actions             = [var.sns_topic_arn]
+  alarm_name                = "${var.prefix}-app-service-action-problem"
+  comparison_operator       = "GreaterThanOrEqualToThreshold"
+  datapoints_to_alarm       = 1
+  evaluation_periods        = 1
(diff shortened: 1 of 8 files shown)


terraform/modules/application/alarms.tf
```diff
@@ -38,6 +38,44 @@ resource "aws_cloudwatch_metric_alarm" "app_memory" {
...
+  alarm_name                = "${var.prefix}-app-tasks-exited"
+  comparison_operator       = "GreaterThanOrEqualToThreshold"
+  datapoints_to_alarm       = 1
+  evaluation_periods        = 1
+  metric_name               = "TriggeredRules"
+  namespace                 = "AWS/Events"
+  ok_actions                = [var.sns_topic_arn]
+  period                    = 8 * 60
+  statistic                 = "Sum"
+  threshold                 = 2 * var.app_task_desired_count
```
SamSeed-Softwire, line 52: After artificially causing this failure for both 2 and 4 tasks, I found that seeing double the desired count fail happened reliably in an 8 minute period.

terraform/modules/application/alarms.tf
```diff
@@ -38,6 +38,44 @@ resource "aws_cloudwatch_metric_alarm" "app_memory" {
...
+  insufficient_data_actions = []
+
+  dimensions = {
+    RuleName = aws_cloudwatch_event_rule.app_task_exited.name
+  }
+}
+
+resource "aws_cloudwatch_metric_alarm" "app_service_action_problem" {
```
SamSeed-Softwire, line 60: I felt that that the conditions that would lead to a WARN or ERROR state were all things we'd want to know about straight away, rather than something we'd want to see happening over a period of time. That's why I went for 1/1/1 for datapoints/periods/threshold here.

The conditions that would lead to a WARN or ERROR state can be seen in the 'Service action events' section [here](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs_cwe_events.html).

SamSeed-Softwire: P.S. I haven't tested this alarm works because I haven't managed to cause the event which would trigger it. I'm happy with the logic though.

terraform/modules/front_door/alarms.tf
```diff
@@ -0,0 +1,19 @@
+resource "aws_cloudwatch_metric_alarm" "unhealthy_hosts_count" {
```
SamSeed-Softwire, line 1: I was quite unsure about datapoints/periods/threshold here. For now I've just put it so that we'll know very quickly if there's any unhealthiness at all (as that seems useful to me!), but lmk if you think we want to monitor over a longer period. E.g. could check for 3 datapoints in 5x 1 minute periods.

terraform/modules/application/alarms.tf
Chirag-Bhatti: Might be worth adding a comment about why an 8min period in the code (in case someone looks at it in the future and is wondering why)

terraform/modules/application/events.tf
```diff
@@ -0,0 +1,43 @@
...
+        ],
+        "stoppedReason" : [
+          "Essential container in task exited"
+        ]
+      }
+    }
+  )
+}
+
+resource "aws_cloudwatch_event_rule" "app_service_action_problem" {
```
Chirag-Bhatti, line 24: Could comment in the stackoverflow post you mentioned to me here, if you think it would be useful for a future viewer of the code

terraform/modules/application/events.tf
```diff
@@ -0,0 +1,43 @@
...
+  event_pattern = jsonencode(
+    {
+      "source" : [
+        "aws.ecs"
+      ],
+      "detail-type" : [
+        "ECS Service Action"
+      ],
+      "resources" : [
+        aws_ecs_service.app.name
```
Chirag-Bhatti, line 36: Looks like this may need the `arn` instead (see [here](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-events-structure.html)), and may be why we haven't seen it triggered it when the ECS was failing to launch tasks y'day during my experimenting

I only got to this link by looking at `event_pattern` terraform docs [here](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_event_rule#event_pattern), which sends you to [here](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-events.html), which then has a link to the above (Event structure reference) 😅

terraform/modules/application/alarms.tf
Chirag-Bhatti: See my comment below about resource `arn` instead of `name` (may help)

terraform/modules/front_door/alarms.tf
Chirag-Bhatti: Yeah think I agree. I believe this is based of healthchecks failing which are likely to happen either because the app/host is genuinely unhealthy, is under so much load the endpoint is not responding / timing out or the AZ has gone down 😱

The healthcheck has criteria itself too before the app/host is deemed unhealthy (see example below), so imo I think it's ok to flag this right away 😊

```
health_check {
    healthy_threshold   = "3"
    interval            = "30"
    protocol            = "HTTP"
    matcher             = "204"
    timeout             = "3"
    path                = "/health"
    unhealthy_threshold = "2"
  }
```

Chirag-Bhatti:
Generally LGTM, left some minor comments and will let Rachael look over it too 👍

terraform/modules/front_door/alarms.tf
RachaelBooth: It's not necessarily clear that we'd want to know about all unhealthy hosts, for the case where one becomes unhealthy and is replaced and then everything continues fine.

I'm happy to stick with this and tweak it later though.

terraform/modules/front_door/alarms.tf
```diff
@@ -0,0 +1,19 @@
...
+  period                    = 60
+  statistic                 = "Maximum"
+  threshold                 = 0
+  insufficient_data_actions = []
+
+  dimensions = {
+    LoadBalancer = aws_lb.this.arn_suffix
+    TargetGroup  = aws_lb_target_group.this.arn_suffix
+  }
+}
```
RachaelBooth, line 19: Do we also want an alarm for not enough healthy hosts? Or will that be sufficiently covered by the others?

RachaelBooth: I think there's an additional state where one (or more) of the targets doesn't exist. E.g. if the service has one task running and it's healthy but there are supposed to be two, then that's still a problem even though there aren't any unhealthy hosts.

I guess that the attempts to start new tasks in that scenario ought to be being caught by the other alarms (my previous attempts at having something like this work in other scenarios was alarm for unhealthy and for too few healthy (and failed deployments), but didn't have anything looking at the ecs events).

SamSeed-Softwire: Ok fair enough, I misunderstood the potential states. I'll add that extra alarm now then 🙂

SamSeed-Softwire: Ok @RachaelBooth this is done now if you're happy to re-review. Unless I'm mistaken, the number of healthy hosts should equal the desired number of tasks, is that right? On that note I guess we will also need to move these alarms into a per-app module when we start the review apps work. I decided not to do that now because I thought it would make sense to do that at the same time we move the target group (as you described [here](https://dluhcdigital.atlassian.net/browse/CLDC-2811)) - especially since we don't know yet whether we're going to put the target group in the application module or a new module.

RachaelBooth (approved):
LGTM. We'll have to see how the thresholds go over time, but I think we can only really find things out by running with them for a bit.

PR #51, opened 2023-09-28
CLDC-2546: fix and remove WAF todos

Chirag-Bhatti:
- Apply WAF protection to load balancer for extra security
- Update WAF rules to mitigate log4j vulnerability


terraform/modules/front_door/waf_cloudfront.tf
Chirag-Bhatti, line 1: So in general I expect us to have the same waf configuration for cloudfront and the load balancer (in terms of web acl and regex patterns). Cloudfront needs the us-east-1 provider and cloudfront scope, whereas load balancer needs eu-west-2 (our default) and regional scope. 

Because of this I wanted to make the WAF a module and set these dynamically, but it turns out you can't set the provider dynamically, so I have had to essentially make two very similar files (but with these bits differing), and add a comment to top of both saying that you should consider updating both when changing any of the WAF configuration 😥. I don't like it but it's the best I could think of, if you know a better idea please let me know!

terraform/modules/front_door/cloudfront.tf
```diff
@@ -4,10 +4,7 @@ locals {
 }
 
 #tfsec:ignore:aws-cloudfront-enable-logging:TODO we will be implementing logging later
-#tfsec:ignore:aws-cloudfront-enable-waf:TODO CLDC-2546
 resource "aws_cloudfront_distribution" "this" {
-  #checkov:skip=CKV2_AWS_47:TODO CLDC-2546 when setting up WAF it should be configured appropriately to mitigate against the Log4j vulnerability https://docs.bridgecrew.io/docs/ensure-aws-cloudfront-attached-wafv2-webacl-is-configured-with-amr-for-log4j-vulnerability
```
Chirag-Bhatti, line 9: This todo had a link to [here](https://docs.bridgecrew.io/docs/ensure-aws-cloudfront-attached-wafv2-webacl-is-configured-with-amr-for-log4j-vulnerability), which suggested adding the rules `AWSManagedRulesKnownBadInputsRuleSet` (which we already have) and `AWSManagedRulesAnonymousIpList` (added below) in our WAF acl to mitigate a log4j vulnerability

`AWSManagedRulesAnonymousIpList` can block requests from VPNs and Proxies (see info [here](https://docs.aws.amazon.com/waf/latest/developerguide/aws-managed-rule-groups-ip-rep.html)) which might cause issues though, so I'll check this works ok locally (since we use GlobalProtect) and we may need to keep an eye on it down the line

terraform/modules/front_door/waf_cloudfront.tf
```diff
@@ -1,6 +1,7 @@
-resource "aws_wafv2_web_acl" "this" {
+# When updating this cloudfront WAF acl, consider making the same change for the load balancer WAF acl, as they generally need to be the same
```
Chirag-Bhatti, line 1: So in general I expect us to have the same waf configuration for cloudfront and the load balancer (in terms of web acl and regex patterns). Cloudfront needs the us-east-1 provider and cloudfront scope, whereas load balancer needs eu-west-2 (our default) and regional scope.

Because of this I wanted to make the WAF a module and set these dynamically, but it turns out you can't set the provider dynamically, so I have had to essentially make two very similar files (but with these bits differing), and add a comment to top of both saying that you should consider updating both when changing any of the WAF configuration 😥. I don't like it but it's the best I could think of, if you know a better idea please let me know!

terraform/modules/front_door/cloudfront.tf
Chirag-Bhatti: This has been applied to staging and my local machine (with GlobalProtect VPN) seems to work OK

RachaelBooth: I'm a bit concerned that end users at local authorities / housing associations that use VPNs may get caught by this too (i.e. it's probably not enough to just know that we can access it while on our VPN). I'm not sure how much you've looked into this - do you have a sense of how likely it is that this becomes a problem?

terraform/modules/front_door/waf_cloudfront.tf
RachaelBooth: What does it gain us to have (the same) waf set up on the load balancer as well as on cloudfront?

Chirag-Bhatti: My main thinking was our LB is public, so if a bad actor was able to send traffic directly there (bypassing cloudfront) then we'd have a similar level of protection at the LB (helping to reduce our overall attack surface on the app). There would be a few other layers to get through to achieve this though (sg's, headers etc), so it's not necessarily a particularly vulnerable area. Was also thinking about pen test outcomes.

I think this implementation implies doubling our WAF costs as it's now analysing the same stuff at two points, so there is a cost/benefit to ponder. I'm happy to remove the LB protection if you feel this trade-off isn't worth it?

RachaelBooth: They would have to get hold of the header we use for preventing accessing the lb directly - which is our intended main source of protection here. 

Presumably as well as cost it also has some impact on performance, which isn't ideal. So I am generally reluctant to be applying it twice.

(N.B. I'm fairly certain delta only applies their waf rules to cloudfront, so fairly happy that it ought to pass a pen test)

terraform/modules/front_door/cloudfront.tf
Chirag-Bhatti: Yeah I agree with the concern 😅. There's nothing concrete from AWS about how much this affects legitimate users of VPNs. The best info I was able to find was the below (not very good):

- A comment on a reddit post saying half their customers were blocked [here](https://www.reddit.com/r/aws/comments/smrxvz/how_to_remove_address_from/)
- A question on AWS re:post clarifying about VPNs [here] (https://repost.aws/questions/QUGXTX6M9RTBS3BQnvIiJFuA/aws-waf-specifically-block-tor)

I suspect you'll want to leave the rule out in this case, and add a code comment about not being able to use it?

terraform/modules/front_door/waf_cloudfront.tf
Chirag-Bhatti: Yeah indeed, and I believe even with the header you'd need to get past the sg's (which checks the source has a cloudwatch prefix) 😀

Good point about the performance! I think I'll remove WAF from LB then all things considered. A lot of the PR will whittle down to just removing and updating the TODO comments rather than any implementation work, but it's good to have considered carefully, thanks 👍

terraform/modules/front_door/cloudfront.tf
RachaelBooth: Drat - you'd really think AWS would be able to be a bit cleverer about it, maybe there isn't really a way to be.

Probably something to leave out in that case.

Chirag-Bhatti: I've removed the rule 😀👍

terraform/modules/front_door/waf_cloudfront.tf
Chirag-Bhatti: I've removed WAF protection on the LB and updated the code comments 😀👍

RachaelBooth (approved):

PR #52, opened 2023-09-28
CLDC-2756: Make redis highly available

SamSeed-Softwire:
https://dluhcdigital.atlassian.net/browse/CLDC-2756

Creating a highly-available (HA) setup with one primary and one replica node is most easily achieved by getting rid of the `aws_elasticache_cluster` resource entirely and using a `aws_elasticache_replication_group` resource instead. However we only want HA for some environments (prod and maybe staging), so I used a flag (at the redis module entrypoint) which is used in `count` expressions in the module to determine which of the resources gets created.

For now I've left staging with a HA setup in case whoever's reviewing this wants to test the failover. I can revert this to non-HA if needed though (the apply may take up to 30 minutes though - seems to take a long time to destroy and recreate these redis resources).

In terms of testing failover:
- This can be done by finding the cluster in the AWS console, selecting the primary node, and then clicking 'Failover primary' (I followed the instructions [here](https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/AutoFailover.html#auto-failover-test) - in particular note that "During the failover process, the console continues to show the node's status as available. To track the progress of your failover test, choose Events from the console navigation pane.")
- When I did this:
  - The app seemed to continue working fine (I was clicking around in it)
  - I checked the app logs and didn't see anything related to the failover
  - I checked the sidekiq logs and saw some connection errors, and then saw messages like this "Redis is online, 35.something sec downtime" (these messages seem to be gone from the logs now, which is weird, but I definitely saw them!)
  - I'm assuming that's a successful failover, but shout if you disagree

### DIFF
--- a/terraform/development/main.tf
+++ b/terraform/development/main.tf
@@ -152,6 +152,7 @@ module "redis" {
 
   prefix                  = local.prefix
   ecs_security_group_id   = module.application.ecs_security_group_id
+  highly_available        = false
   node_type               = "cache.t4g.micro"
   redis_port              = local.redis_port
   redis_subnet_group_name = module.networking.redis_private_subnet_group_name
--- a/terraform/modules/elasticache/outputs.tf
+++ b/terraform/modules/elasticache/outputs.tf
@@ -1,5 +1,5 @@
 output "redis_connection_string" {
-  value       = var.highly_available ? "redis://${aws_elasticache_replication_group.this.primary_endpoint_address}:${var.redis_port}" : "redis://${aws_elasticache_cluster.this.cache_nodes[0].address}:${aws_elasticache_cluster.this.cache_nodes[0].port}"
+  value       = var.highly_available ? "redis://${aws_elasticache_replication_group.this[0].primary_endpoint_address}:${var.redis_port}" : "redis://${aws_elasticache_cluster.this[0].cache_nodes[0].address}:${aws_elasticache_cluster.this[0].cache_nodes[0].port}"
   description = "A connection string for connecting to the primary redis node, intended to be set in the environment variable REDIS_CONFIG"
   sensitive   = true
 }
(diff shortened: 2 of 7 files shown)


terraform/development/main.tf
```diff
@@ -152,6 +152,7 @@ module "redis" {
 
   prefix                  = local.prefix
   ecs_security_group_id   = module.application.ecs_security_group_id
+  highly_available        = false
```
SamSeed-Softwire, line 155: Open to suggestions on a better name for this flag!

terraform/modules/elasticache/outputs.tf
```diff
@@ -1,5 +1,9 @@
 output "redis_connection_string" {
-  value       = "redis://${aws_elasticache_cluster.this.cache_nodes[0].address}:${aws_elasticache_cluster.this.cache_nodes[0].port}"
+  value = var.highly_available ? (
```
SamSeed-Softwire, line 2: Splitting over multiple lines makes this look more messy in some ways, but I think does improve code readability overall.

terraform/modules/elasticache/redis.tf
```diff
@@ -1,11 +1,39 @@
...
+  replication_group_id        = var.prefix
+  security_group_ids          = [aws_security_group.this.id]
+  subnet_group_name           = var.redis_subnet_group_name
+}
+
 #tfsec:ignore:aws-elasticache-enable-backup-retention:TODO CLDC-2679 setup a snapshot retention limit
 resource "aws_elasticache_cluster" "this" {
   #checkov:skip=CKV_AWS_134:TODO CLDC-2679 setup a snapshot retention limit
-  #TODO CLDC-2679 setup redis replicas with multi-az and automatic failover
-  #TODO CLDC-2682 setup redis logging
```
SamSeed-Softwire, line 5: I didn't think having this logging TODO comment was actually helpful. It should be clear what to do when we start that ticket.

RachaelBooth (approved):

PR #53, opened 2023-09-29
CLDC-2755: Make DB highly available

Chirag-Bhatti:
- Implement a single DB read replica (i.e. Multi-AZ DB instance rather than cluster). I agree this is probably what Gov PaaS uses, and one replica is probably sufficient for CORE's requirements
- Implement a flag dictating if the replica will be made in the infra
- Set the flag to true for Prod only (staging is only true for now in case we need to test anything else during review, but will be removed once we're satisfied)
- I tested a DB failover using the instructions [here](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_RebootInstance.html), essentially rebooting our primary DB with a failover. I could see the read replica take over, and a new replica take its place. The app seemed to handle failover fine. I was logged into CORE viewing lettings logs, triggered a failover and tried to load some pages during it. The pages took some time to load (~10's of seconds) but did succeed. The ECS logs also didn't seem to note any app errors talking to the DB. The AWS docs [here](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.MultiAZSingleStandby.html) suggest your app cleans and re-establishes any DB connections, but I suspect ActiveRecord may be handling this already.

### DIFF
--- a/terraform/modules/rds/database.tf
+++ b/terraform/modules/rds/database.tf
@@ -49,14 +49,13 @@ resource "aws_db_instance" "replica" {
   #checkov:skip=CKV2_AWS_30:query logging TODO CLDC-2660
   count               = var.create_replica_standby_db ? 1 : 0
   identifier          = "${var.prefix}-replica"
-  replicate_source_db = aws_db_instance.this.arn
+  replicate_source_db = aws_db_instance.this.identifier
 
   apply_immediately          = aws_db_instance.this.apply_immediately
   auto_minor_version_upgrade = aws_db_instance.this.auto_minor_version_upgrade
   copy_tags_to_snapshot      = aws_db_instance.this.copy_tags_to_snapshot
   db_subnet_group_name       = aws_db_instance.this.db_subnet_group_name
   delete_automated_backups   = aws_db_instance.this.delete_automated_backups
-  deletion_protection        = true # needs to be set to false and applied if you need to delete the DB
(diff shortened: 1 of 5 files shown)


terraform/modules/rds/database.tf
```diff
@@ -38,6 +38,37 @@ resource "aws_db_instance" "this" {
...
+  copy_tags_to_snapshot      = aws_db_instance.this.copy_tags_to_snapshot
+  delete_automated_backups   = aws_db_instance.this.delete_automated_backups
+  deletion_protection        = true
+  final_snapshot_identifier  = aws_db_instance.this.final_snapshot_identifier
+  instance_class             = aws_db_instance.this.instance_class
+  maintenance_window         = aws_db_instance.this.maintenance_window
+  multi_az                   = aws_db_instance.this.multi_az
+  port                       = aws_db_instance.this.port
+  publicly_accessible        = aws_db_instance.this.publicly_accessible
+  replicate_source_db        = aws_db_instance.this.identifier
```
Chirag-Bhatti, line 61: The replica uses the same `aws_db_instance` terraform resource as the primary but also defines a `replicate_source_db`. Some attributes can't / shouldn't be set as per the [docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/db_instance) and things I discovered during implementation, but the attributes that can be set I've opted to ensure are the same as the primary so that we have a true replica 😀

RachaelBooth (approved):
LGTM!

PR #54, opened 2023-09-29
CLDC-2607: Confirm production sizing

RachaelBooth:
This rearranges some settings within the entrypoint modules (for both staging and production) because it made it easier to pull out what might be different and what wouldn't be.

(I've left dev alone because we're going to drastically mess with it soon for review apps)

Couple of changes to numbers of instances in the production module but mostly it already matched what paas seems to have.

### DIFF
--- a/terraform/production/main.tf
+++ b/terraform/production/main.tf
@@ -36,24 +36,38 @@ provider "aws" {
 }
 
 locals {
-  prefix                    = "core-prod"
+  prefix = "core-prod"
+
+  rails_env = "production"
+
   app_host                  = "submit-social-housing-data.levellingup.gov.uk"
-  app_task_desired_count    = 2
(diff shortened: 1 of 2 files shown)


terraform/production/main.tf
```diff
@@ -36,45 +36,54 @@ provider "aws" {
...
 }
 
 module "application" {
   source = "../modules/application"
 
+  app_task_cpu    = 512
+  app_task_memory = 1024
+
+  sidekiq_task_cpu           = 1024
+  sidekiq_task_desired_count = 2
```
RachaelBooth, line 62: This seems slightly pointless, but does protect against an instance going down at the wrong moment I guess? I think it is likely not worth it, just slightly hesitant to be different from the current production.

PR #55, opened 2023-09-29
Production setup

RachaelBooth:
This is replacing the production sizing alone PR because of merge conflicts - so sorry, slightly more to review at once.

On sizing (copied from that PR):

> This rearranges some settings within the entrypoint modules because it made it easier to pull out what might be different and what wouldn't be.
> 
> Couple of changes to numbers of instances in the production module but mostly it already matched what paas seems to have.

On other things:

This sets up an easy way to restrict access by ip, but doesn't yet put it in place for any environment (mainly because I want the DLUHC ones to work, particularly at the point when they are making DNS changes).

This also sets up an easy way + instructions to create the things necessary to be able to tell DLUHC what DNS records are needed without getting problems because of certs not being validated yet as things that use them are created, so making it relatively painless to set up a new environment.


terraform/modules/application_security_group/main.tf
RachaelBooth, line 1: Pulling this module out means that the front door does not depend on the application, which makes creating just the things needed for DNS records a lot easier (and it's something that has to happen for the review apps work).

terraform/modules/front_door/ip_list.tf
```diff
@@ -0,0 +1,13 @@
...
+  # This is used when restrict_by_ip is set to true, to restrict access to testing environments
+  ip_allowlist = [
+    # Softwire
+    "31.221.86.178",
+    "167.98.33.82",
+    "82.163.115.98",
+    "87.224.105.250",
+    "87.224.116.242/32",
+    "45.150.142.210/32"
+    # DLUHC - TODO
```
RachaelBooth, line 11: We'll need to ask DLUHC for these. I'm content to merge and add them later if we get to that point though.

terraform/modules/front_door/load_balancer.tf
```diff
@@ -51,8 +53,15 @@ resource "aws_lb_listener" "https" {
   }
 }
 
+moved {
```
RachaelBooth, line 56: This and the other moved blocks are here for clarity for the moment - they can be removed once we've done an apply on all existing environments (which I now have) but I was going to leave them in for the review process so it's clear what's happened.

terraform/production/main.tf
```diff
@@ -36,47 +36,64 @@ provider "aws" {
...
 }
 
 module "application" {
   source = "../modules/application"
 
+  app_task_cpu    = 512
+  app_task_memory = 1024
+
+  sidekiq_task_cpu           = 1024
+  sidekiq_task_desired_count = 2
```
RachaelBooth, line 62: This seems a bit pointless, but is what's happening on paas. I suppose it protects against a task going down at just the wrong moment. Nevertheless we could probably get rid of it, but my default was to match paas.

SamSeed-Softwire (approved):
Lgtm, apart from removing the `moved` blocks of course.

PR #56, opened 2023-10-02
CLDC-2786: Sidekiq alarms

SamSeed-Softwire:
I couldn't create healthy/unhealthy hosts alarms for sidekiq because there's no target group. I think it's fine not to have these given we're not expecting sidekiq tasks to fail anywhere near as much as the app tasks (correct me if I'm wrong!)

### DIFF
--- a/terraform/modules/application/alarms.tf
+++ b/terraform/modules/application/alarms.tf
@@ -157,7 +157,7 @@ resource "aws_cloudwatch_metric_alarm" "sidekiq_service_action_problem" {
 resource "aws_cloudwatch_metric_alarm" "sidekiq_running_tasks" {
   alarm_actions             = [var.sns_topic_arn]
   alarm_name                = "${var.prefix}-sidekiq-running-tasks"
-  comparison_operator       = "LessOrEqualToThanThreshold"
+  comparison_operator       = "LessThanOrEqualToThreshold"
   datapoints_to_alarm       = 1
   evaluation_periods        = 1
   metric_name               = "RunningTaskCount"
--- a/terraform/modules/application/alarms.tf
+++ b/terraform/modules/application/alarms.tf
@@ -115,3 +115,41 @@ resource "aws_cloudwatch_metric_alarm" "sidekiq_memory" {
     ServiceName = aws_ecs_service.sidekiq.name
   }
 }
+
+resource "aws_cloudwatch_metric_alarm" "sidekiq_tasks_exited" {
+  alarm_actions             = [var.sns_topic_arn]
+  alarm_name                = "${var.prefix}-sidekiq-tasks-exited"
+  comparison_operator       = "GreaterThanOrEqualToThreshold"
+  datapoints_to_alarm       = 1
+  evaluation_periods        = 1
+  metric_name               = "TriggeredRules"
+  namespace                 = "AWS/Events"
+  ok_actions                = [var.sns_topic_arn]
+  period                    = 3 * 60 # By causing sidekiq to fail repeatedly (with both 1 tasks and 2 tasks running) I found that double the desired count would reliably fail in a 3 minute period.
+  statistic                 = "Sum"
+  threshold                 = 2 * var.sidekiq_task_desired_count
+  insufficient_data_actions = []
+
+  dimensions = {
+    RuleName = aws_cloudwatch_event_rule.sidekiq_task_exited.name
+  }
+}
+
+resource "aws_cloudwatch_metric_alarm" "sidekiq_service_action_problem" {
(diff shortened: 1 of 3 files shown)


terraform/modules/application/alarms.tf
```diff
@@ -115,3 +115,41 @@ resource "aws_cloudwatch_metric_alarm" "sidekiq_memory" {
...
+resource "aws_cloudwatch_metric_alarm" "sidekiq_tasks_exited" {
+  alarm_actions             = [var.sns_topic_arn]
+  alarm_name                = "${var.prefix}-sidekiq-tasks-exited"
+  comparison_operator       = "GreaterThanOrEqualToThreshold"
+  datapoints_to_alarm       = 1
+  evaluation_periods        = 1
+  metric_name               = "TriggeredRules"
+  namespace                 = "AWS/Events"
+  ok_actions                = [var.sns_topic_arn]
+  period                    = 3 * 60 # By causing sidekiq to fail repeatedly (with both 1 tasks and 2 tasks running) I found that double the desired count would reliably fail in a 3 minute period.
```
SamSeed-Softwire, line 128: In order to test this I updated the command in the container definition to be this: `command = ["echo", "'Testing failure!'", "&&", "exit", "1"]`. (I then triggered a deployment from the core repo (branch `ForceAwsTaskFailure`) to ensure the service was pulling a valid image.)

SamSeed-Softwire: Here are the container starts with desired_count = 1

(I stopped this after about 10 minutes)
image

SamSeed-Softwire: Here are the container starts with desired_count = 2. I also used this to decide on the period.

(I also stopped this after about 10 minutes)
image

RachaelBooth (approved):
This looks good (just slight question on confidence and whether we think we need something counting running tasks).

SamSeed-Softwire:
> > I couldn't create healthy/unhealthy hosts alarms for sidekiq because there's no target group. I think it's fine not to have these given we're not expecting sidekiq tasks to fail anywhere near as much as the app tasks (correct me if I'm wrong!)
> 
> I don't think it's frequency we're concerned about so much as whether this is sufficient to ensure we spot any problems.
> 
> It is less of an issue if sidekiq goes down without us noticing than for the main app, but it's also more likely that it could not be functioning without us noticing.
> 
> So it depends if we're confident this will catch problems or if we think we need to have something that's on a "is a thing running" basis. You've actually looked into the events more than I have, do you have a sense on that?

Fair enough. I looked into this and found there is a `RunningTaskCount` metric that can be used, it just requires enabling container insights on the cluster (which I've now done, and created a sensible alarm). I don't think this undermines the other alarms I've done so far - I think `RunningTaskCount` could potentially be used to replace the healthy hosts count alarm (but there's no need to replace it), and I think all the other alarms I've got are useful for other reasons. What do you think?

terraform/modules/application/ecs.tf
```diff
@@ -2,6 +2,11 @@
 resource "aws_ecs_cluster" "this" {
   #checkov:skip=CKV_AWS_65:TODO CLDC-2542 enable container insights if necessary for logging/monitoring
   name = var.prefix
+
+  setting {
+    name  = "containerInsights"
+    value = "enabled"
+  }
```
RachaelBooth, line 9: FYI: Container insights is relatively expensive, because it sends a number of metrics per cluster / per task family / per service to cloudwatch which are all charged as custom metrics (the basic ones automatically sent by AWS services are not charged for). This is why AWS tells you about the workaround using a count on a CPU metric. Looking at it at the moment on staging it's sending 63 metrics, which I think works out at ~$18.90 per month.

We're probably ok with that (certainly it's dwarfed by the cost of dev time), but it's worth being aware of with this kind of thing.

terraform/modules/application/ecs.tf
```diff
@@ -2,6 +2,11 @@
 resource "aws_ecs_cluster" "this" {
   #checkov:skip=CKV_AWS_65:TODO CLDC-2542 enable container insights if necessary for logging/monitoring
```
RachaelBooth, line 3: If we are using container insights now we can get rid of this skip.

terraform/modules/application/ecs.tf
SamSeed-Softwire: @RachaelBooth I've switched to the cpu count approach to save money - only took me 10 mins so worth the time I think! Lmk if you're happy with the new metric :) Thanks!

RachaelBooth: LGTM

PR #57, opened 2023-10-02
CLDC-2687: Update apply_immediately for rds & redis

SamSeed-Softwire:
Whether or not to apply changes immediately to rds and redis has been pulled out into a variable for each of those modules (as we want to apply changes immediately for some environments but not others).

### DIFF
--- a/terraform/staging/main.tf
+++ b/terraform/staging/main.tf
@@ -126,11 +126,11 @@ module "certificates" {
 module "database" {
   source = "../modules/rds"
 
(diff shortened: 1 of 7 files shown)


terraform/staging/main.tf
SamSeed-Softwire, line 1: I thought we'd probably still want to apply changes immediately for staging. Shout if you disagree.

RachaelBooth (approved):

PR #58, opened 2023-10-03
CLDC-2548: Configure shield

SamSeed-Softwire:
Implement basic shield protections.

I've raised a ticket to implement additional shield protections in the future: https://dluhcdigital.atlassian.net/browse/CLDC-2857.

### DIFF
--- a/terraform/modules/front_door/shield.tf
+++ b/terraform/modules/front_door/shield.tf
@@ -4,9 +4,21 @@ resource "aws_shield_protection" "cloudfront" {
   resource_arn = aws_cloudfront_distribution.this.arn
 }
 
+resource "aws_shield_application_layer_automatic_response" "cloudfront" {
+  count        = var.enable_aws_shield ? 1 : 0
+  action       = "BLOCK"
+  resource_arn = aws_cloudfront_distribution.this.arn
+}
+
 resource "aws_shield_protection" "load_balancer" {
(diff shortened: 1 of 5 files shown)


terraform/production/main.tf
```diff
@@ -47,6 +47,8 @@ locals {
 
   app_task_desired_count = 4
 
+  enable_aws_shield = false
```
SamSeed-Softwire, line 50: Currently want this disabled everywhere until we get the green light.

terraform/modules/front_door/shield.tf
SamSeed-Softwire, line 1: In the future if we add other shield protections not specific to our front door (e.g. the policy which allows the Shield Response Team to access our account) then we may want to move all the shield resources out into a separate module. I think it's fine here for now though.

SamSeed-Softwire, line 1: In the future if we add other shield protections not specific to our front door (e.g. the policy which allows the Shield Response Team to access our account) then we may want to move all the shield resources out into a separate module. I think it's fine here for now though.

On the other hand if you think that having lots of count statements is messy, I'd understand if you wanted to move it out into a separate module now!

RachaelBooth (commented):
LGTM once the checks are sorted

RachaelBooth (approved):

PR #62, opened 2023-10-04
Updates for production environment creation

RachaelBooth:


terraform/modules/cds_export/bucket.tf
```diff
@@ -8,7 +8,7 @@ resource "aws_s3_bucket" "this" {
   #checkov:skip=CKV_AWS_145: default encryption is fine
   #checkov:skip=CKV_AWS_144: cross region replication is overkill when this is only for data transfer
   #checkov:skip=CKV_AWS_21: versioning not important, data source is elsewhere
-  bucket = "${var.prefix}-export"
+  bucket = "${var.prefix}-export${var.bucket_suffix}"
```
RachaelBooth, line 11: Annoyingly this caused a duplicate name error without an extra suffix - I didn't want to change the prefix for this module, so the optional suffix seemed to be a reasonable answer.

terraform/modules/front_door/ip_list.tf
```diff
@@ -2,10 +2,10 @@ locals {
...
   ip_allowlist = [
     # Softwire
-    "31.221.86.178",
-    "167.98.33.82",
-    "82.163.115.98",
-    "87.224.105.250",
+    "31.221.86.178/32",
+    "167.98.33.82/32",
+    "82.163.115.98/32",
+    "87.224.105.250/32",
```
SamSeed-Softwire, line 8: If we have Cambridge and Manchester do we also want to add Bucharest here? (I found the rest of the above here https://softwiretech.atlassian.net/wiki/spaces/Helpdesk/pages/6856573750/Network+WAN+Administration.)

terraform/modules/front_door/waf.tf
```diff
@@ -285,7 +285,6 @@ resource "aws_wafv2_regex_pattern_set" "waf_rate_limit_urls" {
 
 resource "aws_wafv2_ip_set" "allowed_ips" {
   provider = aws.us-east-1
-  count    = var.restrict_by_ip ? 1 : 0
```
SamSeed-Softwire, line 288: Why would we no longer want this?

SamSeed-Softwire (approved):
Lgtm subject to the couple of questions I've asked. I know we need to add DLUHC IPs but probably best to merge this for now and raise another PR once we have them?

terraform/modules/front_door/waf.tf
RachaelBooth: In this case, mainly because deleting it when turning the restriction off is annoying (mainly for ordering reasons - which we could sort out differently but it was easier not to for now). Will come back to this when sorting out which envs the ip restriction is on for and the remaining ips.

PR #66, opened 2023-10-09
CLDC-2534: ECR image scanning notifications

Chirag-Bhatti:
- Implement event rule which triggers when an ECR image scan finds MEDIUM or higher severity vulnerabilities
- Test with manually created email subscriber (instead of Slack) for now
- TODO ensure the event rule publishes notification to Slack channel. This will be possible once AWS Chatbot in [CLDC-2544 (Setup notifications to slack)](https://dluhcdigital.atlassian.net/browse/CLDC-2544) has been resolved


Chirag-Bhatti:
I've been able to confirm the event was triggered in AWS console and an email was sent as below 👍 
image

terraform/modules/ecr/events.tf
```diff
@@ -0,0 +1,38 @@
+resource "aws_cloudwatch_event_rule" "ecr_image_scan_results" {
+  name        = "${var.prefix}-ecr-image-scan-results"
```
RachaelBooth, line 2: This might be better referencing the name of the repository rather than a chosen env prefix.

RachaelBooth (approved):
One naming comment, I'm happy for that to be adjusted without me seeing it again.

PR #67, opened 2023-10-09
CLDC-2811: refactor application secrets into separate module

Chirag-Bhatti:
- Refactored secrets used in application module into a separate module


terraform/development/main.tf
```diff
@@ -80,13 +81,46 @@ module "application" {
...
+  os_data_key_secret_arn               = module.application_secrets.os_data_key_secret_arn
   private_subnet_ids                   = module.networking.private_subnet_ids
   rails_env                            = local.rails_env
+  rails_master_key_secret_arn          = module.application_secrets.rails_master_key_secret_arn
   redis_connection_string              = module.redis.redis_connection_string
+  sentry_dsn_secret_arn                = module.application_secrets.sentry_dsn_secret_arn
   sns_topic_arn                        = module.monitoring.sns_topic_arn
 }
 
+moved {
```
RachaelBooth, line 95: These shouldn't be needed for dev, although they shouldn't cause errors by being here.

RachaelBooth (approved):

terraform/development/main.tf
Chirag-Bhatti: Good point! I'll remove to save any confusion

PR #68, opened 2023-10-10
CLDC-2811: Refactor task, task execution and deployment roles

Chirag-Bhatti:
Refactor the task, task execution and deployment roles into a separate module


terraform/staging/main.tf
```diff
@@ -66,30 +66,121 @@ module "application" {
...
   os_data_key_secret_arn               = module.application_secrets.os_data_key_secret_arn
-  private_subnet_ids                   = module.networking.private_subnet_ids
-  rails_env                            = local.rails_env
   rails_master_key_secret_arn          = module.application_secrets.rails_master_key_secret_arn
-  redis_connection_string              = module.redis.redis_connection_string
   sentry_dsn_secret_arn                = module.application_secrets.sentry_dsn_secret_arn
-  sns_topic_arn                        = module.monitoring.sns_topic_arn
 }
 
 moved {
```
SamSeed-Softwire, line 186: Are these moved statements being left in for now because they haven't yet been applied to production? Presumably they've already been done for staging?

terraform/production/main.tf
```diff
@@ -66,30 +66,107 @@ module "application" {
...
   sidekiq_task_memory        = 8192
 
-  ecr_repository_url      = "815624722760.dkr.ecr.eu-west-2.amazonaws.com/core"
+  ecr_repository_url = "815624722760.dkr.ecr.eu-west-2.amazonaws.com/core"
+
+  prefix                          = local.prefix
+  api_key_secret_arn              = module.application_secrets.api_key_secret_arn
+  app_host                        = local.app_host
+  app_task_desired_count          = local.app_task_desired_count
+  app_task_execution_role_arn     = module.application_roles.ecs_task_execution_role_arn
```
SamSeed-Softwire, line 75: Do we actually want to prefix `app_task_execution_role_arn` and `app_task_role_arn` with `app`, given they're being used by all 3 tasks?

SamSeed-Softwire (commented):
Lgtm!

terraform/staging/main.tf
Chirag-Bhatti: Yep they should definitely remain for production as that won't be applied yet.

I could technically remove these from `staging/main.tf` now as I have applied it, but they're a useful record in case I need to move something back / elsewhere. My plan is to remove all these blocks at the end once the whole ticket is deemed OK 😀

terraform/production/main.tf
Chirag-Bhatti: Good spot! I think these should be `ecs_task_`... instead of `app_task_` so have made that change 👍

terraform/modules/application_roles/variables.tf
```diff
@@ -0,0 +1,54 @@
+variable "api_key_secret_arn" {
```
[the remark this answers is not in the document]
Chirag-Bhatti: Yeah that's a good idea, I've made the change 👍

RachaelBooth (approved):

PR #69, opened 2023-10-11
CLDC-2811: Refactor load balancer target group and listener rule

Chirag-Bhatti:
Refactor the load balancer target group and listener rule from the `front_door` to the `application` module


terraform/modules/application/load_balancer.tf
```diff
@@ -0,0 +1,36 @@
...
+    path                = "/health"
+    unhealthy_threshold = "2"
+  }
+}
+
+resource "aws_lb_listener_rule" "forward_cloudfront" {
+  count = var.initial_create ? 0 : 1
+
+  listener_arn = var.load_balancer_listener_arn
+  priority     = 1
```
RachaelBooth, line 23: I think in order to allow multiple applications we have to not set a priority value here (meaning it should be given the highest available priority on creation), otherwise we'll have a clash.

terraform/modules/application/load_balancer.tf
```diff
@@ -0,0 +1,36 @@
...
+    protocol            = "HTTP"
+    matcher             = "204"
+    timeout             = "3"
+    path                = "/health"
+    unhealthy_threshold = "2"
+  }
+}
+
+resource "aws_lb_listener_rule" "forward_cloudfront" {
+  count = var.initial_create ? 0 : 1
```
RachaelBooth, line 20: I don't think this is necessary: on an initial creation, we shouldn't be creating resources in the application module at all - unless something in this PR has changed that, but if so I'm worried that that means the review app setup won't work.

terraform/modules/front_door/alarms.tf
```diff
@@ -14,7 +14,7 @@ resource "aws_cloudwatch_metric_alarm" "healthy_hosts_count" {
 
   dimensions = {
     LoadBalancer = aws_lb.this.arn_suffix
-    TargetGroup  = aws_lb_target_group.this.arn_suffix
+    TargetGroup  = var.load_balancer_target_group_arn_suffix
```
RachaelBooth, line 17: This must not have existed when the JIRA ticket was written else it would be referenced explicitly there, but it will need pulling out of this module in order to meet the criteria that no shared module references anything from a per-app module.

RachaelBooth: This may already be on your list for later! But it seemed worth commenting on because the variable is new.

RachaelBooth (approved):

PR #70, opened 2023-10-11
CLDC-2785 Fixing email issues

SamSeed-Softwire:


RachaelBooth (approved):
This all looks good - do we actually want to also add the other env vars/secrets? (I'm mainly thinking that it wouldn't be obvious that if you added something new to the app that sidekiq did and the app already did that you might need to add an env var/secret here)

SamSeed-Softwire:
> This all looks good - do we actually want to also add the other env vars/secrets? (I'm mainly thinking that it wouldn't be obvious that if you added something new to the app that sidekiq did and the app already did that you might need to add an env var/secret here)

Thanks @RachaelBooth. Yeah that sounds like a sensible thing to do! Although presumably that would very slightly worsen our security profile as the container wouldn't have the minimum it needed to function? But in practical terms this is presumably fine - the other containers have access to all of the secrets and we are not storing any secrets in an extra location. 

Will implement now and merge.

PR #71, opened 2023-10-12
CLDC-2811: Refactor DB name and connection string

Chirag-Bhatti:
Refactor the DB name and connection string across the application and RDS modules instead of just the RDS module


terraform/development/main.tf
```diff
@@ -52,8 +54,6 @@ locals {
   application_port = 8080
   database_port    = 5432
   redis_port       = 6379
-
-  create_db_migration_infra = false
```
Chirag-Bhatti, line 56: We don't need this in dev so removed

terraform/development/main.tf
```diff
@@ -69,27 +69,28 @@ module "application" {
...
-  sns_topic_arn                        = module.monitoring.sns_topic_arn
+  prefix                                 = local.prefix
+  api_key_secret_arn                     = module.application_secrets.api_key_secret_arn
+  app_host                               = local.app_host
+  app_task_desired_count                 = local.app_task_desired_count
+  application_port                       = local.application_port
+  bulk_upload_bucket_access_policy_arn   = module.bulk_upload.read_write_policy_arn
+  bulk_upload_bucket_details             = module.bulk_upload.details
+  database_data_access_policy_arn        = module.database.rds_data_access_policy_arn
+  database_name                          = "${local.default_database_name}-${terraform.workspace}"
```
Chirag-Bhatti, line 80: Just pre-empting a possible `database_name` for review apps here, happy to change or do something different

terraform/development/main.tf
```diff
@@ -150,27 +151,6 @@ module "database" {
   vpc_id                = module.networking.vpc_id
 }
 
-module "database_migration" {
```
Chirag-Bhatti, line 153: We don't need this in dev so removed the module

terraform/modules/application/parameters.tf
```diff
@@ -0,0 +1,10 @@
+data "aws_ssm_parameter" "partial_database_connection_string" {
```
Chirag-Bhatti, line 1: I was unsure if this was OK / there's a better way of building the complete database connecting string in the application module. 

My concern was whether this might expose the partial database connecting string to the state file (which contains the db endpoint, username and password)? It might be encrypted however so not directly useful? Any info / thoughts on this appreciated 😀

terraform/modules/rds/database.tf
```diff
@@ -68,10 +67,3 @@ resource "aws_db_instance" "replica" {
     prevent_destroy = true
   }
 }
-
-resource "aws_ssm_parameter" "database_connection_string" {
```
Chirag-Bhatti, line 72: Thought it was clearer to move this into a separate `parameters.tf` file (see `terraform/modules/rds/parameters.tf`) while adapting it

terraform/modules/application/parameters.tf
Chirag-Bhatti: I haven't applied this to staging yet until we're sure what it will / won't expose

terraform/modules/rds/parameters.tf
```diff
@@ -0,0 +1,6 @@
+resource "aws_ssm_parameter" "database_partial_connection_string" {
+  #checkov:skip=CKV_AWS_337:default encryption not using a kms cmk sufficient
+  name  = "DATABASE_PARTIAL_CONNECTION_STRING"
```
Chirag-Bhatti, line 3: Renaming this from `DATA_COLLECTOR_DATABASE_URL` makes terraform want to recreate it (which I've generally been trying to avoid), but I think it's ok in this case

terraform/production/main.tf
```diff
@@ -185,20 +193,23 @@ module "database_migration" {
...
+  db_migration_task_memory   = 16384
+  ecs_task_ephemeral_storage = 200 #GiB
+
+  ecr_repository_url = "815624722760.dkr.ecr.eu-west-2.amazonaws.com/db-migration"
+
+  cloudfoundry_service = "dluhc-core-production-postgres"
+  cloudfoundry_space   = "production"
+
+  prefix                                  = local.prefix
+  database_complete_connection_string_arn = module.application.rds_complete_connection_string_arn
```
Chirag-Bhatti, line 206: This is the only real change here (likewise for the other envs), replacing the `database_connection_string_arn`. I tidied up the variable here so it looks worse than it is 😀.

terraform/staging/main.tf
```diff
@@ -69,27 +71,28 @@ module "application" {
...
+  prefix                                 = local.prefix
+  api_key_secret_arn                     = module.application_secrets.api_key_secret_arn
+  app_host                               = local.app_host
+  app_task_desired_count                 = local.app_task_desired_count
+  application_port                       = local.application_port
+  bulk_upload_bucket_access_policy_arn   = module.bulk_upload.read_write_policy_arn
+  bulk_upload_bucket_details             = module.bulk_upload.details
+  database_data_access_policy_arn        = module.database.rds_data_access_policy_arn
+  database_name                          = local.default_database_name
+  database_partial_connection_string_arn = module.database.rds_partial_connection_string_arn
```
Chirag-Bhatti, line 83: We now pass in this variable instead of `database_connection_string_arn` for all the envs. And then build the complete connection string in the application module. 

Hopefully the concept I created of passing a "partial" connection string and building it into a "complete" one in this PR is clear for the code reader

terraform/modules/application/parameters.tf
RachaelBooth: I don't think this changes anything about what ends up in the terraform state - the db password, endpoint etc will already be there (but are not useful unless you can first get access to somewhere the db will accept connections from).

terraform/modules/rds/parameters.tf
RachaelBooth: Yeah, seems fine to let it do that.

terraform/modules/application/parameters.tf
```diff
@@ -0,0 +1,10 @@
+data "aws_ssm_parameter" "partial_database_connection_string" {
+  name = var.database_partial_connection_string_arn
+}
+
+resource "aws_ssm_parameter" "complete_database_connection_string" {
+  #checkov:skip=CKV_AWS_337:default encryption not using a kms cmk sufficient
+  name  = "COMPLETE_DATABASE_CONNECTION_STRING"
```
RachaelBooth, line 7: This name may need to have the prefix in, so that it's unique across different app instances.

terraform/modules/application/parameters.tf
Chirag-Bhatti: Thanks for explaining, I'll leave it as is 😊

RachaelBooth (approved):

PR #73, opened 2023-10-13
CLDC-2909: Restrict http access to bulk upload and cds export buckets

Chirag-Bhatti:
- Restrict the ability to make http connections with the our `bulk upload` and `cds export` buckets

I've checked the state and state replica buckets in the meta account, they both already have this enabled (through the external modules)

### DIFF
--- a/terraform/modules/bulk_upload/bucket.tf
+++ b/terraform/modules/bulk_upload/bucket.tf
@@ -25,7 +25,7 @@ resource "aws_s3_bucket_policy" "force_ssl" {
   bucket = aws_s3_bucket.this.id
 
   policy = jsonencode({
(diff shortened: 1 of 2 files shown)


terraform/modules/bulk_upload/bucket.tf
```diff
@@ -20,6 +20,32 @@ resource "aws_s3_bucket_public_access_block" "this" {
   restrict_public_buckets = true
 }
 
+
```
Chirag-Bhatti, line 23: The policy has been duplicated here across the two files, but it seemed better than creating a module with just a bucket policy. I think if we end up duplicating it more than twice we should modularise

RachaelBooth (approved):

PR #74, opened 2023-10-13
CLDC-2908: Implement access logging for bulk upload and cds export buckets

Chirag-Bhatti:
- Enable access logging on the bulk upload and cds export buckets, with the logs going to the separate dedicated log buckets
- Check the access logs appear in the dedicated buckets ✅

### DIFF
--- a/terraform/modules/bulk_upload/access_log_bucket.tf
+++ b/terraform/modules/bulk_upload/access_log_bucket.tf
@@ -1,10 +1,10 @@
-#tfsec:ignore:aws-s3-enable-bucket-encryption: default bucket encryption is sufficient
+#tfsec:ignore:aws-s3-enable-bucket-encryption: access log buckets not compatible with kms encryption, see https://docs.aws.amazon.com/AmazonS3/latest/userguide/troubleshooting-server-access-logging.html
 #tfsec:ignore:aws-s3-enable-bucket-logging: access log bucket doesn't need access logging itself
-#tfsec:ignore:aws-s3-encryption-customer-key: default encryption is sufficient
+#tfsec:ignore:aws-s3-encryption-customer-key: access log buckets are not compatible with kms encryption, see https://docs.aws.amazon.com/AmazonS3/latest/userguide/troubleshooting-server-access-logging.html
 resource "aws_s3_bucket" "bulk_upload_access_logs" {
   #checkov:skip=CKV2_AWS_62: no need for event notifications
   #checkov:skip=CKV_AWS_18: access log bucket doesn't need access logging itself
-  #checkov:skip=CKV_AWS_145: default encryption is fine
+  #checkov:skip=CKV_AWS_145: access log buckets are not compatible with kms encryption, see https://docs.aws.amazon.com/AmazonS3/latest/userguide/troubleshooting-server-access-logging.html
   #checkov:skip=CKV_AWS_144: cross region replication not required for access logs
   bucket = "${var.prefix}-bulk-upload-logs"
 }
--- a/terraform/modules/bulk_upload/access_log_bucket.tf
+++ b/terraform/modules/bulk_upload/access_log_bucket.tf
@@ -29,23 +29,25 @@ resource "aws_s3_bucket_policy" "allow_access_logs" {
     Version = "2012-10-17",
     Statement = [
       {
-        Sid       = "S3ServerAccessLogsPolicy",
-        Action    = "s3:PutObject",
-        Effect    = "Allow",
-        Principal = "logging.s3.amazonaws.com",
+        Sid    = "S3ServerAccessLogsPolicy",
+        Action = "s3:PutObject",
+        Effect = "Allow",
+        Principal = {
+          "Service" = "logging.s3.amazonaws.com"
+        },
         Resource = [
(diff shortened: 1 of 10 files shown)


terraform/staging/main.tf
```diff
@@ -135,12 +135,42 @@ module "application_security_group" {
   vpc_id                          = module.networking.vpc_id
 }
 
+moved {
```
Chirag-Bhatti, line 138: Plan to remove the moved blocks after PR has been approved and applied to staging (production ones should remain)

Chirag-Bhatti:
We're starting to see access logs in the dedicated buckets 😀, here's an example: 
image

terraform/modules/bulk_upload/access_log_bucket.tf
```diff
@@ -0,0 +1,72 @@
+#tfsec:ignore:aws-s3-enable-bucket-encryption: default bucket encryption is sufficient
+#tfsec:ignore:aws-s3-enable-bucket-logging: access log bucket doesn't need access logging itself
+#tfsec:ignore:aws-s3-encryption-customer-key: default encryption is sufficient
+#tfsec:ignore:aws-s3-enable-versioning: Not important, each log will be a new file with a different name
```
RachaelBooth, line 4: Do we want versioning as a form of deletion protection?

terraform/modules/bulk_upload/access_log_bucket.tf
```diff
@@ -0,0 +1,72 @@
...
+resource "aws_s3_bucket_lifecycle_configuration" "bulk_upload_access_logs" {
+  bucket = aws_s3_bucket.bulk_upload_access_logs.id
+
+  rule {
+    id = "expire-old-logs"
+
+    filter {}
+
+    expiration {
+      days = 90
```
RachaelBooth, line 63: We might need to consider this length (given that I think one of the low level issues raised was that we should keep cloudwatch logs for much longer), but I think we could decide to leave it for now and consider them all in one go.

terraform/modules/bulk_upload/access_log_bucket.tf
Chirag-Bhatti: I think we do, implemented in this PR now 👍

terraform/modules/bulk_upload/access_log_bucket.tf
Chirag-Bhatti: Yeah ideally better to do in one go, is there a ticket for this?

terraform/modules/bulk_upload/access_log_bucket.tf
```diff
@@ -0,0 +1,101 @@
+#tfsec:ignore:aws-s3-enable-bucket-encryption: default bucket encryption is sufficient
```
SamSeed-Softwire, line 1: Do we want to update the explanations for this (and also in line 3) because KMS is actually not supported for access logging (as commented [here](https://dluhcdigital.atlassian.net/jira/software/c/projects/CLDC/boards/115?selectedIssue=CLDC-2917&sprints=619))?

terraform/modules/bulk_upload/access_log_bucket.tf
```diff
@@ -0,0 +1,101 @@
+#tfsec:ignore:aws-s3-enable-bucket-encryption: default bucket encryption is sufficient
+#tfsec:ignore:aws-s3-enable-bucket-logging: access log bucket doesn't need access logging itself
+#tfsec:ignore:aws-s3-encryption-customer-key: default encryption is sufficient
+resource "aws_s3_bucket" "bulk_upload_access_logs" {
```
SamSeed-Softwire, line 4: Why name the resource `_access_logs` but the bucket itself just `_logs`?

terraform/modules/cds_export/access_log_bucket.tf
```diff
@@ -0,0 +1,101 @@
...
+
+    filter {}
+
+    noncurrent_version_expiration {
+      noncurrent_days = 90
+    }
+
+    noncurrent_version_transition {
+      noncurrent_days = 30
+      storage_class   = "STANDARD_IA"
```
Chirag-Bhatti, line 98: Since we're doing bucket versioning, it seemed prudent to move non-current versions of log files into a cheaper storage class in case they hang around. I picked `STANDARD_IA` as it seemed the best balance of cost / availability, but you could probably argue a case for all the storage classes listed [here](https://aws.amazon.com/s3/storage-classes/)!

terraform/modules/bulk_upload/access_log_bucket.tf
Chirag-Bhatti: Nice spot, yes I think we should so have updated it now 😊

terraform/modules/bulk_upload/access_log_bucket.tf
Chirag-Bhatti: `_logs` follows the convention we set for the backend log bucket names, but I felt `access_logs` made things clearer in the terraform itself since we're doing the implementation ourselves here (rather than an external module as per the backend)

RachaelBooth (approved):

PR #77, opened 2023-10-16
CLDC-2915: Implement load balancer access logging

Chirag-Bhatti:
- Enable access logging on the load balancer, with the logs going to a separate dedicated log bucket
- Check the access logs appear in the dedicated bucket ✅

### DIFF
--- a/terraform/modules/front_door/access_log_bucket.tf
+++ b/terraform/modules/front_door/access_log_bucket.tf
@@ -1,0 +1,91 @@
+#tfsec:ignore:aws-s3-enable-bucket-encryption: access log buckets not compatible with kms encryption, see https://docs.aws.amazon.com/AmazonS3/latest/userguide/troubleshooting-server-access-logging.html
+#tfsec:ignore:aws-s3-enable-bucket-logging: access log bucket doesn't need access logging itself
+#tfsec:ignore:aws-s3-encryption-customer-key: access log buckets not compatible with kms encryption, see https://docs.aws.amazon.com/AmazonS3/latest/userguide/troubleshooting-server-access-logging.html
+resource "aws_s3_bucket" "load_balancer_access_logs" {
+  #checkov:skip=CKV2_AWS_62: no need for event notifications
+  #checkov:skip=CKV_AWS_18: access log bucket doesn't need access logging itself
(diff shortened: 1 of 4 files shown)


terraform/modules/front_door/access_log_bucket.tf
```diff
@@ -0,0 +1,91 @@
...
+        Condition = {
+          Bool = {
+            "aws:SecureTransport" = "false"
+          }
+        },
+        Sid    = "AllowLoadBalancerAccessLogs",
+        Action = "s3:PutObject",
+        Effect = "Allow",
+        Principal = {
+          AWS = data.aws_elb_service_account.current.arn
```
Chirag-Bhatti, line 47: Here AWS says you need to put the **account ID of the AWS region** your load balancer is in (not the account ID you've made the load balancer in itself), see "Attach a policy to your S3 bucket - Regions available before Aug 2022" section [here](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/enable-access-logging.html)

The terraform above is a programmatic way of getting this information rather than hardcoding it 😀

Chirag-Bhatti:
I spotted an issue with `terraform apply` regarding these changes. It often won't apply successfully first time round but after a few mins will the second time round. I've placed some more info Confluence [here](https://dluhcdigital.atlassian.net/wiki/spaces/MC/pages/31064299/Destroying+recreating+infra) in the LB section

Chirag-Bhatti:
Here's a screenshot of the load balancer logs going into the bucket 😀

image

terraform/modules/front_door/load_balancer_access_log_bucket.tf
```diff
@@ -20,7 +20,7 @@ resource "aws_s3_bucket_public_access_block" "load_balancer_access_logs" {
 
 data "aws_elb_service_account" "current" {}
 
-resource "aws_s3_bucket_policy" "force_ssl_and_allow_access_logs" {
+resource "aws_s3_bucket_policy" "force_ssl_and_allow_load_balancer_access_logs" {
```
Chirag-Bhatti, line 23: I'm anticipating a potential clash with an upcoming Cloudfront access logging PR, so have renamed this file and this bucket policy to avoid it (since there will be two sets of similar resources in the front door module, one for cloudfront and one for load balancer access logging)

RachaelBooth (approved):

PR #78, opened 2023-10-17
CLDC-2911: Implement cloudfront access logging

Chirag-Bhatti:
- Enable access logging on our cloudfront distribution, with the logs going to a separate dedicated log bucket
- Check the access logs appear in the dedicated bucket


terraform/modules/front_door/cloudfront.tf
```diff
@@ -59,6 +60,12 @@ resource "aws_cloudfront_distribution" "this" {
     }
   }
 
+  logging_config {
+    bucket          = aws_s3_bucket.cloudfront_access_logs.bucket_domain_name
+    include_cookies = false
```
Chirag-Bhatti, line 65: Do you agree there isn't value in recording cookie information in the access logs?

If enabled, Cloudfront will log all cookies in the access logs regardless of it they are usually forward to the origin or not. I believe we use a `_data_collector_session` cookie in the app, and in the browser it looks like this stores an encrypted value about the session. I didn't personally see much value in logging this and other analytics cookies but may have missed something!

terraform/modules/front_door/cloudfront_access_log_bucket.tf
```diff
@@ -0,0 +1,104 @@
...
+        Resource = [
+          aws_s3_bucket.cloudfront_access_logs.arn,
+          "${aws_s3_bucket.cloudfront_access_logs.arn}/*"
+        ],
+        Condition = {
+          Bool = {
+            "aws:SecureTransport" = "false"
+          }
+        },
+        Sid    = "AllowCloudfrontAccessLogs",
```
Chirag-Bhatti, line 51: Haven't been able to prove if this is bit of the bucket policy is required, since Cloudfront uses bucket ACLs to enable it to write to the log bucket (I've seen contradicting information on AWS docs however about this policy 😀). I may remove it in a later commit when I do some further testing. Right now I know the implementation works regardless.

Chirag-Bhatti:
Here's a screenshot of cloudfront access logs appearing in the bucket and their content 😊

image
image

terraform/modules/front_door/cloudfront_access_log_bucket.tf
Chirag-Bhatti: I've proven the cloudwatch access logs work without this policy, so have now removed 👍

Chirag-Bhatti:
For further info, it looks as though Cloudwatch access logs are output at every 5mins or slower (I guess it skips a 5min interval if there are no logs to put out). I can't see that this is configurable anywhere, just that it's the default behaviour presumably but I think that this is probably OK.

image

terraform/modules/front_door/cloudfront.tf
```diff
@@ -59,6 +60,12 @@ resource "aws_cloudfront_distribution" "this" {
...
+    prefix          = ""
```
SamSeed-Softwire, line 66: `prefix` is optional according to the terraform docs - do we need this line?

terraform/modules/front_door/cloudfront_access_log_bucket.tf
```diff
@@ -0,0 +1,90 @@
...
+# See link for info: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AccessLogs.html#access-logs-choosing-s3-bucket
+resource "aws_s3_bucket_ownership_controls" "cloudfront_access_logs" {
+  bucket = aws_s3_bucket.cloudfront_access_logs.id
+
+  rule {
+    object_ownership = "BucketOwnerPreferred"
+  }
+}
+
+resource "aws_s3_bucket_policy" "force_ssl_and_allow_cloudfront_access_logs" {
```
SamSeed-Softwire, line 31: For the load balancer we had a `AllowLoadBalancerAccessLogs` statement which referenced `data.aws_elb_service_account.current.arn`. We don't have an equivalent statement here - as such should the name of this policy really include `_and_allow_cloudfront_access_logs`?

SamSeed-Softwire (commented):
Lgtm, subject to my comments.

terraform/modules/front_door/cloudfront.tf
SamSeed-Softwire: Sounds reasonable to me, but Rachael might have more knowledge to bring to the question!

terraform/modules/front_door/cloudfront_access_log_bucket.tf
Chirag-Bhatti: Actually in my comment above I removed this bit of the bucket policy (allowing logs from a particular AWS service to be put in the bucket) because here we use bucket ACLs. So I should change this to something like `force_ssl_cloudfront_access_logs`. I'll update, nice find 👍

Chirag-Bhatti: Updated to just `force_ssl` 👍

terraform/modules/front_door/cloudfront.tf
RachaelBooth: Would the cookies ever give us useful information if we needed to look through the logs? I'm not immediately sure of the answer.

terraform/modules/front_door/cloudfront.tf
Chirag-Bhatti: For our `tf_state_backend` terraform module and `aws_s3_bucket_logging` resources, we had to provide similar prefixes with an empty string 😅, so I wanted to be consistent and explicit here as well, with the benefit that it might be clear the logs live in the root of the bucket

terraform/modules/front_door/cloudfront.tf
Chirag-Bhatti: Had a chat with Kat about this

Looks like the `_data_collection_session` cookie holds your sign-in info and it's value is sensitive (because someone who had access to where this info is stored could then sign in as you, by just setting it as their own cookie), so not sure it's a good idea to record it. It also stores info like filters selected on the page which I don't think is useful. I believe the Cloudwatch Access Logs will most likely be viewed if there is a security concern, and we need to verify what requests have come in and from which Client IPs for example, so I'm leaning strongly on it isn't necessary and risky to store if we do.

RachaelBooth (approved):

PR #80, opened 2023-10-18
Customer-managed KMS

SamSeed-Softwire:
This PR covers using customer-managed KMS keys for the following:
* S3 buckets - https://dluhcdigital.atlassian.net/browse/CLDC-2917
* SSM parameters - https://dluhcdigital.atlassian.net/browse/CLDC-2918
* Secrets Manager secrets - https://dluhcdigital.atlassian.net/browse/CLDC-2919

I think it could potentially be refactored/simplified, but wanted to get this raised for discussion first. Some points worth mentioning on that topic:
* I felt it was cleaner to keep all KMS stuff within the module it was relevant to. So instead of creating a new KMS module, I added a `kms.tf` file in all the modules that needed a CMK.
* I ended up creating one KMS key per module - that felt like an appropriate balance between encryption isolation and complexity. However if we want to actually have one KMS key per encrypted resource then I can definitely do that, in which case I'd probably rename each key to match the name of the resource it's encrypting. (Alternatively if we want only one KMS key for the whole infrastructure I can also do that, which would probably mean creating a dedicated KMS module.)
* In all the KMS key policies I've got a statement which enables the account root user to do all KMS actions. When I just had the statement which allowed a particular role (task or task_execution) to perform the GenerateDataKey and Decrypt actions I got an error when applying the terraform - it said I wouldn't be able to manage the keys after creation. I felt that assigning the root user all KMS permissions seemed like an appropriate response to this. AFAICT we don't actually need this statement in all the key policies - just in one of them - but picking one arbitrarily didn't feel like a clean solution. We could also create a dedicated KMS module just to contain that policy, but that feels like overkill. Wanted to discuss this :)

Will `fmt` once approved.


terraform/modules/application_secrets/kms.tf
```diff
@@ -0,0 +1,52 @@
...
+      type        = "AWS"
+      identifiers = ["arn:aws:iam::${data.aws_caller_identity.current.account_id}:root"]
+    }
+
+    actions = [
+      "kms:*"
+    ]
+
+    resources = [
+      "*"
```
RachaelBooth, line 49: Does this need to be on everything or could it be on this specific kms key? (if so that would avoid some of the checkov skips, and the repetition, although I agree we're going to want this for every key)

terraform/modules/application/parameters.tf
```diff
@@ -3,7 +3,7 @@ data "aws_ssm_parameter" "partial_database_connection_string" {
 }
 
 resource "aws_ssm_parameter" "complete_database_connection_string" {
-  #checkov:skip=CKV_AWS_337:default encryption not using a kms cmk sufficient
+  key_id = aws_kms_key.this.arn
```
RachaelBooth, line 6: Can we keep name at the top here? (I think this is coming from required vs non-required parameters, or what's more immediately going to want to be known to a human reading this - or maybe just that standard terraform examples always start with name?)

terraform/modules/application/kms.tf
```diff
@@ -0,0 +1,52 @@
...
+  statement {
+    principals {
+      type        = "AWS"
+      identifiers = [var.ecs_task_execution_role_arn]
+    }
+
+    actions = [
+      "kms:GenerateDataKey",
+      "kms:Decrypt"
+    ]
```
RachaelBooth, line 31: Are both permissions needed here?

RachaelBooth: Actually to be fair in this case the role will definitely need to decrypt, and I'm less bothered about whether the generate data key permission is technically needed - but it's still worth thinking about.

terraform/modules/cds_export/kms.tf
```diff
@@ -0,0 +1,50 @@
...
+}
+
+data "aws_iam_policy_document" "kms" {
+  #checkov:skip=CKV_AWS_109:Only assigning the kms:GenerateDataKey and kms:Decrypt permissions led to a 'you won't be able to manage the key once created' error
+  #checkov:skip=CKV_AWS_111:Only assigning the kms:GenerateDataKey and kms:Decrypt permissions led to a 'you won't be able to manage the key once created' error
+  #checkov:skip=CKV_AWS_356:Only assigning the kms:GenerateDataKey and kms:Decrypt permissions led to a 'you won't be able to manage the key once created' error
+  statement {
+    principals {
+      type        = "AWS"
+      identifiers = [var.ecs_task_role_arn]
```
RachaelBooth, line 23: Here, won't the task role only need to be able to encrypt, while the cds role will need to be able to decrypt?

terraform/modules/application_secrets/kms.tf
Chirag-Bhatti: Tested this out and looks like it can just be on the specific kms key, I'll update across the codebase 👍

terraform/modules/application/parameters.tf
Chirag-Bhatti: Will update this here and a few other places where we have a similar thing 👍

terraform/modules/cds_export/kms.tf
Chirag-Bhatti: Yeah I think I agree, the CDS role is also optionally created so this will need to be a separate policy document, that is only applied if the role exists. Will update 👍

terraform/modules/application_secrets/kms.tf
Chirag-Bhatti: Done 👍

terraform/modules/application/parameters.tf
Chirag-Bhatti: Done 👍

terraform/modules/cds_export/kms.tf
Chirag-Bhatti: Done 👍

terraform/modules/application/kms.tf
Chirag-Bhatti: I believe `kms:GenerateDataKey` is generally needed when you're doing client-side decryption / encryption, which I don't believe we are (the S3 buckets use `SSE-KMS` and some looking up on param store / secrets suggests it's server side here too though it isn't obvious)

I'll remove them and see if we break anything in terms of S3 / Param Store / Secrets for the app 😅

Chirag-Bhatti: Removed 👍 , over the weekend some export jobs will try to run which will give some ideas, and I'll redeploy the app to ECS to see if it can still pull down params / secrets

terraform/modules/cds_export/kms.tf
Chirag-Bhatti: I had some issues getting the CDS role policy _and_ the ECS task role to combine and apply to the key (when required), but this has now been fixed by this [commit](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data-infrastructure/pull/80/commits/1fade63b3daecc65b646f600754a7dd70ddad946). I don't love the implementation but it's the only solution I've managed to get working on the key!

terraform/modules/application/kms.tf
Chirag-Bhatti: I tested the theory out by removing the `kms:GenerateDataKey` everywhere, but it broke the Export job and Bulk Upload process. The rest of the app seemed to behave OK but I felt it was better to retain this action everywhere it was originally removed from to avoid quietly breaking something, balanced with fact this action shouldn't cause concern if it is there for being overly permissive 👍

RachaelBooth: That's not surprising - I would expect `kms:GenerateDataKey` to be needed by the task role for the keys used for the bulk upload and export buckets, because the app puts things into those buckets using that role. I don't expect anything using the task execution role to putting data into ssm parameters though - I think this is a separate thing to look at.

Chirag-Bhatti: Removed the `kms:GenerateDataKey` from everywhere that isn't to do with S3 (which meant anything to do with KMS and the task execution role), redeployed the app (it was able to pull down secrets and parameters) and satisfied everything looks to be working fine without it 👍

PR #81, opened 2023-10-18
CLDC-2912: Enable redis backups

SamSeed-Softwire:


terraform/modules/elasticache/redis.tf
```diff
@@ -22,12 +22,12 @@ resource "aws_elasticache_replication_group" "this" {
   preferred_cache_cluster_azs = ["eu-west-2a", "eu-west-2b"] # The first AZ in the list is where the primary node will be created. Replicas will be created in the following AZs.
   replication_group_id        = var.prefix
   security_group_ids          = [aws_security_group.this.id]
+  snapshot_retention_limit    = 5
```
SamSeed-Softwire, line 25: I went for 5 as it seemed to be a standard length used by others. If everything in redis will be short-term cached data which doesn't need storing for more than a day (for example) then I can definitely reduce this down to 1.

RachaelBooth (approved):

PR #83, opened 2023-10-20
CLDC-2597:  Define dev / review apps infrastructure

Chirag-Bhatti:
- Split our dev infrastructure into shared and per-review-app modules
- Update our terraform pipeline accounting for the new structure
- Update any configuration and sizing to make dev suitable for review apps


terraform/modules/front_door/outputs.tf
```diff
@@ -9,7 +9,7 @@ output "load_balancer_arn_suffix" {
 }
 
 output "load_balancer_listener_arn" {
-  value       = aws_lb_listener.https[0].arn
+  value       = var.initial_create ? null : aws_lb_listener.https[0].arn
```
Chirag-Bhatti, line 12: When initially creating infra, this output errored so I added this conditional

terraform/development/shared/main.tf
```diff
@@ -0,0 +1,168 @@
...
+
+  cloudfront_domain_name    = local.app_host
+  load_balancer_domain_name = local.load_balancer_domain_name
+}
+
+module "database" {
+  source = "../../modules/rds"
+
+  allocated_storage       = 100
+  backup_retention_period = 0 # no backups
```
Chirag-Bhatti, line 115: I don't think we need to backup the DB for dev / review-apps

terraform/development/shared/main.tf
```diff
@@ -0,0 +1,168 @@
...
+
+  apply_changes_immediately          = true
+  enable_primary_deletion_protection = false
+  highly_available                   = false
+  skip_final_snapshot                = true
```
Chirag-Bhatti, line 120: For dev / review-apps I don't think we need high availability or a final snapshot on deletion. Thinking over I would enable deletion protection though, because we only have the one cluster not several for each review app so will update this

terraform/development/per_review_app/main.tf
```diff
@@ -0,0 +1,114 @@
...
+  vpc_id                                            = data.terraform_remote_state.development.outputs.networking_vpc_id
+}
+
+module "redis" {
+  source = "../../modules/elasticache"
+
+  snapshot_retention_limit = 5
+
+  apply_changes_immediately = true
+  highly_available          = false
```
Chirag-Bhatti, line 107: I don't think Redis needs to be highly available for review apps

terraform/development/shared/main.tf
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  cloudfront_domain_name    = local.app_host
+  load_balancer_domain_name = local.load_balancer_domain_name
+}
+
+module "database" {
+  source = "../../modules/rds"
+
+  allocated_storage       = 100
```
Chirag-Bhatti, line 114: In PaaS we allocated a DB instance with 5Gb of stage per review app.

In our case we will one DB instance with several DB's inside per review app, so 100Gb felt reasonable to accommodate multiple (~20) review apps at any one time

terraform/development/shared/main.tf
```diff
@@ -0,0 +1,168 @@
...
+  backup_retention_period = 0 # no backups
+
+  apply_changes_immediately          = true
+  enable_primary_deletion_protection = false
+  highly_available                   = false
+  skip_final_snapshot                = true
+  instance_class                     = "db.m5.xlarge"
```
Chirag-Bhatti, line 121: I've gone for this instance class (which is fairly well resourced since we need to accommodate all review app DBs into one DB cluster. See comparisons below between PaaS and AWS. I felt that because review apps won't be exercised all at the same time, something with double the computational power and network performance, and a good amount of RAM should be enough for some consecutive review apps being used and talking to the DBs.

**Gov PaaS (several DB instances, specs per instance):**
- tiny-unencrypted-13 (which is db.t3.micro in AWS)
- 5Gb storage
- 1Gb RAM
- 2 vCPUs / 1 Physical Core
- 5Gbps Network Performance

**AWS (one DB instance to handle all)**
- db.m5.xlarge
- 100Gb storage
- 16Gb RAM
- 4 vCPUs / 2 Physical Cores
- 10Gbps Network Performance

terraform/development/shared/main.tf
RachaelBooth: It occurs to me - have you had a look at the actual amount of data typically stored in review apps (or the free space metrics)? If that's a generous amount we may not need 20x the amount we've been giving a single app.

terraform/modules/front_door/outputs.tf
RachaelBooth: Huh, I think I just saw it as blank before rather than erroring. 🤷‍♀️

terraform/staging/main.tf
```diff
@@ -259,8 +262,10 @@ module "monitoring" {
 module "redis" {
   source = "../modules/elasticache"
 
+  snapshot_retention_limit = 5
+
   apply_changes_immediately = true
-  highly_available          = true
+  highly_available          = false
```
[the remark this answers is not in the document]
RachaelBooth: I suspect it was needed when developing for that ticket, tbh - something we should watch out for when tidying up after new features if they vary by env.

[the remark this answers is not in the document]
Chirag-Bhatti: Yeah I thought I had changed this by accident, but it looks like the Redis on PaaS staging is also _not_ highly available, so my change here should be kept

terraform/development/shared/main.tf
Chirag-Bhatti: On average review apps tend to use ~2.5Gb (half of the storage space available to them, though some breathing space for storage is good, typically at least 10%)

PaaS currently has 22 DBs, but there's only 9 PRs open for the app

**So I'll set this to 50Gb here** (~16/17 apps, each taking 2.5Gb + 20% breathing space => 3Gb each)

---

If we want to be safe we could also configure `max_allocated_storage` in dev, enables autoscaling of the DB storage (without downtime) for the following conditions:
- Free available space is <= 10% of the allocated storage
- The low-storage condition lasts at least 5mins
- At least 6hrs have passed since the last storage modification, or storage optimization has completed on the instance, whichever is longer

The additional storage is in increments of whichever of the following is greater:
- 10 Gb
- 10% of currently allocated storage
- Predicted storage growth exceeding the current allocated storage size in the next 7hrs based on the FreeStorageSpace metrics from the past hour (I'm worried that a burst of review apps over a short time period might unnecessarily trigger this though, it probably works better for gradually reducing free storage space behaviour)

terraform/development/per_review_app/main.tf
```diff
@@ -0,0 +1,114 @@
...
+provider "aws" {
+  region = "eu-west-2"
+
+  assume_role {
+    role_arn = local.provider_role_arn
+  }
+}
+
+locals {
+  prefix = "core-${terraform.workspace}" # terraform workspaces are expected to have a name in the format "review-XXXX"
```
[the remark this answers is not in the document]
Chirag-Bhatti: Updated so that workspace is just a number 👍

terraform/development/per_review_app/main.tf
```diff
@@ -0,0 +1,114 @@
...
+
+  prefix                                            = local.prefix
+  api_key_secret_arn                                = data.terraform_remote_state.development.outputs.application_secrets_api_key_secret_arn
+  app_host                                          = local.app_host
+  app_task_desired_count                            = local.app_task_desired_count
+  application_port                                  = local.application_port
+  bulk_upload_bucket_details                        = data.terraform_remote_state.development.outputs.bulk_upload_details
+  cloudfront_header_name                            = data.terraform_remote_state.development.outputs.front_door_cloudfront_header_name
+  cloudfront_header_password                        = data.terraform_remote_state.development.outputs.front_door_cloudfront_header_password
+  database_name                                     = "${local.prefix}-${local.default_database_name}"
```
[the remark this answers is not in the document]
Chirag-Bhatti: Dropped `data_collector` from the end 👍

terraform/development/per_review_app/main.tf
```diff
@@ -0,0 +1,114 @@
...
+  redis_connection_string                           = module.redis.redis_connection_string
+  sentry_dsn_secret_arn                             = data.terraform_remote_state.development.outputs.application_secrets_sentry_dsn_secret_arn
+  sns_topic_arn                                     = data.terraform_remote_state.development.outputs.monitoring_sns_topic_arn
+  vpc_id                                            = data.terraform_remote_state.development.outputs.networking_vpc_id
+}
+
+module "redis" {
+  source = "../../modules/elasticache"
+
+  snapshot_retention_limit = 5
```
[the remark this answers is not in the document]
Chirag-Bhatti: Changed this to `0` (removed redis backups) and added a checkov ignore comment in the `per_review_app` `main.tf` entrypoint. The comment had to go here rather than inside the elasticache module, to avoid it falsely affecting the other envs

terraform/development/shared/main.tf
Chirag-Bhatti: Enabled deletion protection 👍

terraform/development/shared/main.tf
Chirag-Bhatti: Allocated storage now 50Gb. No autoscaling but can update if you think it's necessary 👍

terraform/modules/elasticache/redis.tf
```diff
@@ -25,24 +22,3 @@ resource "aws_elasticache_replication_group" "this" {
   subnet_group_name           = var.redis_subnet_group_name
   transit_encryption_enabled  = true
 }
-
-resource "aws_elasticache_cluster" "this" {
```
Chirag-Bhatti, line 29: We plan to set `highly_available` for Redis to `false` on staging, but I noticed that meant we'd use the `aws_elasticache_cluster` resource for creating Redis instead of `aws_elasticache_replication_group`, and this would mean losing encryption at rest as it can't be configured on there. 

You can use the replication_group resource for non-high and high availability, so I've updated this here to ensure we retain the encryption. [See this issue](https://github.com/hashicorp/terraform-provider-aws/issues/22123) for more info.

RachaelBooth (approved):

terraform/development/shared/main.tf
```diff
@@ -0,0 +1,168 @@
...
+      version = "~>5.0"
+      source  = "hashicorp/aws"
+    }
+  }
+
+  backend "s3" {
+    bucket         = "core-non-prod-tf-state"
+    dynamodb_table = "core-non-prod-tf-state-lock"
+    encrypt        = true
+    key            = "core-development.tfstate"
```
SamSeed-Softwire, line 15: Do we perhaps want to call this core-shared instead?

SamSeed-Softwire (commented):
Lgtm subject to fixing the checkov. Nice one!

terraform/development/per_review_app/main.tf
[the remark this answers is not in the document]
Chirag-Bhatti: Turns out this doesn't work in the way I thought, it will still ignore it for all envs, so I've had to move it into the elasticache module now 😃

terraform/development/shared/main.tf
Chirag-Bhatti: Would have been nice to but as there's some state tracking already in `core-development.tfstate`, we'd have to migrate the state to `core-shared.tfstate` I think, which is pretty annoying. Especially as the state being tracked currently has what we created to enable us to inform the Dev DNS changes required, so probably not worth it imo 😀

SamSeed-Softwire: Doesn't sound too hard (see [here](https://stackoverflow.com/questions/63464775/terraform-renaming-state-file)) and if done correctly shouldn't affect any of the actual infra AFAICT. But happy to leave as is (or perhaps change later) given it's not critical path.

Chirag-Bhatti: I've managed to make these changes and migrate the state 🎉, thanks!

We now have `core-development-shared.tfstate` and `core-development-per-review-app.tfstate`

terraform/modules/elasticache/redis.tf
```diff
@@ -1,15 +1,17 @@
...
   #checkov:skip=CKV_AWS_31:TODO CLDC-2937 potentially introduce an auth token later
+  #checkov:skip=CKV_AWS_134:redis backups not required for dev / review apps
   #checkov:skip=CKV_AWS_191:default encryption key is sufficient
+  #checkov:skip=CKV2_AWS_50:redis multi-az automatic failover cannot be enabled when we want one node / no high availability
   apply_immediately           = var.apply_changes_immediately
   at_rest_encryption_enabled  = true
   auto_minor_version_upgrade  = true
-  automatic_failover_enabled  = true
-  description                 = "Redis replication group, containing a primary node and a replica."
+  automatic_failover_enabled  = var.highly_available ? true : false
```
RachaelBooth, line 9: The changes to this file look like they ought to have been in a different PR, but it's not worth separating it now.

PR #84, opened 2023-10-24
CLDC-2907: Remove permissions on default security groups

SamSeed-Softwire:
This task involved 2 parts:
- Explicitly referencing the default VPC security groups
  - There are 4 of these: 3 for eu-west-x and 1 for us-east-1 (these were the VPCs called out in the pen test matrix, but we already knew that these are the only regions which we have permission to create resources in)
- "Adopting" into our terraform state the default security group for each of these VPCs, as well as the non-default VPC we have been using so far in the project.

### DIFF
--- a/terraform/development/main.tf
+++ b/terraform/development/main.tf
@@ -27,6 +27,24 @@ provider "aws" {
 }
 
 provider "aws" {
+  alias  = "eu-west-1"
+  region = "eu-west-1"
+
+  assume_role {
+    role_arn = local.provider_role_arn
+  }
+}
+
+provider "aws" {
+  alias  = "eu-west-3"
+  region = "eu-west-3"
+
+  assume_role {
+    role_arn = local.provider_role_arn
+  }
+}
+
+provider "aws" {
   alias  = "us-east-1"
   region = "us-east-1"
 
@@ -198,6 +216,12 @@ module "front_door" {
 
 module "networking" {
   source = "../modules/networking"
+
+  providers = {
+    aws.eu-west-1 = aws.eu-west-1
+    aws.eu-west-3 = aws.eu-west-3
+    aws.us-east-1 = aws.us-east-1
+  }
 
   prefix                                  = local.prefix
   vpc_cidr_block                          = "10.0.0.0/16"
--- a/terraform/modules/networking/security_groups.tf
+++ b/terraform/modules/networking/security_groups.tf
@@ -1,0 +1,21 @@
+resource "aws_default_security_group" "default-eu-west-2" {
+  vpc_id = data.aws_vpc.default-eu-west-2.id
+}
+
+resource "aws_default_security_group" "default-eu-west-1" {
+  provider = aws.eu-west-1
+
+  vpc_id = data.aws_vpc.default-eu-west-1.id
+}
+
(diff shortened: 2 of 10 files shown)


terraform/development/main.tf
```diff
@@ -26,6 +26,24 @@ provider "aws" {
   }
 }
 
+provider "aws" {
```
SamSeed-Softwire, line 29: These aren't in alphameric order because I thought it made sense to keep our default provider at the top.

terraform/modules/networking/security_groups.tf
```diff
@@ -0,0 +1,25 @@
+resource "aws_default_security_group" "main" {
+  vpc_id = aws_vpc.main.id
+}
+
+resource "aws_default_security_group" "default_eu_west_2" {
```
SamSeed-Softwire, line 5: Again I felt it was better to keep the default provider (eu-west-2) at the top, even though this isn't in alphanumeric order.

terraform/modules/networking/vpc.tf
```diff
@@ -84,3 +84,25 @@ data "aws_iam_policy_document" "vpc_flow_logs_assume_role_permissions" {
     ]
   }
 }
+
+data "aws_vpc" "default_eu_west_2" {
+  default = true
+}
+
+data "aws_vpc" "default_eu_west_1" {
```
SamSeed-Softwire, line 92: I tried to define a local variable `non_default_providers = [aws.eu-west-1, aws.eu-west-3, aws.us-east-1]` and then use a single `aws_vpc` data block for these 3 non-default providers (using `count = length(local.non_default_providers)`) but this didn't work - Terraform complained about the use of a `default` attribute when I tried this. So I've left as separate data blocks, which I think is fine.

terraform/staging/main.tf
```diff
@@ -244,11 +262,22 @@ module "front_door" {
...
+    aws.eu-west-3 = aws.eu-west-3
+    aws.us-east-1 = aws.us-east-1
+  }
+
   prefix                                  = local.prefix
   vpc_cidr_block                          = "10.0.0.0/16"
   vpc_flow_cloudwatch_log_expiration_days = 60
 }
 
+moved {
```
SamSeed-Softwire, line 276: I'll remove this at the point of merging, I just wanted to make clear that I'd done this (it's already been applied).

terraform/modules/networking/vpc.tf
RachaelBooth: Do you mean count there or do you mean a for_each looping over the non default providers?

RachaelBooth:
Interestingly, I'm assuming that there are indeed default vpcs and security groups in the other regions, but neither we nor the pen testers can interact with them.

Since we can't interact with them though, because we can't use the regions, we also can't create infra in there that would accidentally use them, so I guess that mitigates any risk.

terraform/modules/networking/vpc.tf
SamSeed-Softwire: I had a go with `for_each` but couldn't get it working, and in the end concluded that it's not possible to do this (based on  some google searches) because terraform treats providers in a different way to regular resources/attributes.

RachaelBooth (approved):

PR #85, opened 2023-10-25
CLDC-NONE: upgrade staging infra temporarily to match production

Chirag-Bhatti:
- Raised for visibility
- Upgrade the staging infra to temporarily match production, enabling us to test db migration and performance
- These changes / commits can be reverted once testing has been completed

### DIFF
--- a/terraform/staging/main.tf
+++ b/terraform/staging/main.tf
@@ -47,7 +47,7 @@ locals {
 
   provider_role_arn = "arn:aws:iam::107155005276:role/developer"
 
-  app_task_desired_count = 2
+  app_task_desired_count = 4
 
   enable_aws_shield = false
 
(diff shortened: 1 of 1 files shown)


terraform/staging/main.tf
```diff
@@ -184,11 +184,11 @@ module "certificates" {
 module "database" {
   source = "../modules/rds"
 
-  allocated_storage         = 25
+  allocated_storage         = 100
   apply_changes_immediately = true
   backup_retention_period   = 7
   highly_available          = false
-  instance_class            = "db.t3.micro"
+  instance_class            = "db.t3.small"
```
Chirag-Bhatti, line 191: We might retain this instance class for staging down the line if the cost save is negligible

RachaelBooth (approved):

PR #86, opened 2023-10-25
Run s3 sync on AWS

RachaelBooth:


RachaelBooth:
N.B. This is currently only wired up for staging, and the bucket names there are actually for review (for testing proof of concept). Those will need updating to actually use.

terraform/modules/s3_migration/kms.tf
```diff
@@ -0,0 +1,29 @@
+resource "aws_kms_key" "this" {
+  description         = "KMS key used for the Secrets Manager secrets."
+  enable_key_rotation = true
+}
+
+resource "aws_kms_alias" "this" {
+  name          = "alias/${var.prefix}-db-migration-secretsmanager-secrets"
```
Chirag-Bhatti, line 7: Should be `...s3-migration` rather than `...db-migration`

RachaelBooth: Ooh, good point I missed that one.

SamSeed-Softwire (commented):
Lgtm but I will avoid approving for now in case we run into any issues with it when doing the migration today.

terraform/staging/main.tf
```diff
@@ -55,7 +55,9 @@ locals {
   database_port    = 5432
   redis_port       = 6379
 
+  # separate for testing
```
SamSeed-Softwire, line 58: Do we still want this comment?

terraform/staging/main.tf
```diff
@@ -224,6 +226,30 @@ module "database_migration" {
...
+  source = "../modules/s3_migration"
+
+  count = local.create_s3_migration_infra ? 1 : 0
+
+  ecr_repository_url = "815624722760.dkr.ecr.eu-west-2.amazonaws.com/s3-migration"
+
+  prefix = local.prefix
+  buckets = {
+    export = {
+      source      = "s3://paas-s3-broker-prod-lon-791e541d-6996-44c8-9884-322448062d7b",
```
SamSeed-Softwire, line 239: I've updated these bucket urls to be correct for staging.

SamSeed-Softwire (approved):
Lgtm, only a couple of minor comments.

PR #87, opened 2023-10-26
Set up ip restrictions

RachaelBooth:


terraform/modules/front_door/ip_list.tf
```diff
@@ -7,7 +7,12 @@ locals {
     "82.163.115.98/32",
     "87.224.105.250/32",
     "87.224.116.242/32",
-    "45.150.142.210/32"
-    # DLUHC - TODO
+    "45.150.142.210/32",
+    # DLUHC
+    "35.176.187.166/32"
+  ]
+  ip_allowlist_test = [
```
RachaelBooth, line 14: I'm a bit suspicious of the need for this, but it is what DLUHC are telling us so lets go for it and can adjust it later if need be.

terraform/modules/front_door/variables.tf
```diff
@@ -53,6 +53,12 @@ variable "restrict_by_ip" {
   description = "True if access to cloudfront should be restricted by ip, e.g. before release or for a test environment"
 }
 
+variable "include_test_ips" {
+  type        = bool
+  default     = true
```
RachaelBooth, line 58: It's this way round with the default because it made slightly more sense given that post switchover everything that remains restricted will want to include test things.

Chirag-Bhatti: Minor thing - the variable needs to move up in the file in order to keep things in alphabetical order 😀

Chirag-Bhatti (approved):
LGTM - just one minor comment. Happy for it to be merged once resolved

terraform/modules/front_door/variables.tf
RachaelBooth: Ah, of course - I put it next to the thing it was linked to while writing. Obviously I should have come up with a name that would keep it there?

RachaelBooth: Have actually thought of such a name because it makes way more sense to me to have these variables next to each other.

PR #88, opened 2023-10-26
CLDC-2619: migrate staging data changes

Chirag-Bhatti:
- Fixes required for successfully migrating the staging DB
- Setting create db migration infra to true for staging


db-migration/run.sh
```diff
@@ -11,7 +11,7 @@ fi
 
 echo "Setting up conduit into Gov PaaS DB and making dump file"
 
-if cf conduit ${CF_SERVICE} -- pg_dump -v -j 2 --file dumpfile --no-acl --encoding utf8 --clean --no-owner --if-exists -Fc; then
+if cf conduit ${CF_SERVICE} -- pg_dump -v --file dumpfile --no-acl --encoding utf8 --clean --no-owner --if-exists -Fc; then
```
Chirag-Bhatti, line 14: Can't use workers with the -Fc option (which is an option to use the custom file format for pg_dump)

db-migration/run.sh
```diff
@@ -2,7 +2,7 @@
 
 echo 'Logging into Gov PaaS'
 
-if cf login -a api.london.cloud.service.gov.uk -u ${CF_USERNAME} -p ${CF_PASSWORD} -s ${CF_SPACE}; then
+if cf login -a api.london.cloud.service.gov.uk -o "dluhc-core" -u ${CF_USERNAME} -p ${CF_PASSWORD} -s ${CF_SPACE}; then
```
Chirag-Bhatti, line 5: Don't necessarily need the org here (now that we realised it was that the logon didn't have access to the staging space causing the issue, not this) but there's no harm either

terraform/modules/ecr_migration/main.tf
```diff
@@ -9,6 +9,7 @@ terraform {
   }
 }
 
+#tfsec:ignore:aws-ecr-enforce-immutable-repository:mutable images preferred
```
RachaelBooth, line 12: This looks like a duplicate of two lines below?

Chirag-Bhatti: Whoops good spot! I think I missed this when I merged the latest changes in from main (about s3-migration). Removed now 😀

RachaelBooth (approved):

PR #89, opened 2023-10-27
CLDC-NONE: remove unrequired tf moved blocks in prod

Chirag-Bhatti:
- Removal of moved blocks in prod which are incorrect / not needed


terraform/production/main.tf
```diff
@@ -275,11 +270,6 @@ module "application_secrets" {
   ecs_task_execution_role_arn = module.application_roles.ecs_task_execution_role_arn
 }
 
-moved {
-  from = module.redis.aws_security_group.this
-  to   = module.application_roles.aws_security_group.redis
-}
-
```
Chirag-Bhatti, line 282: There's another moved block (on line 261) which is the correct one

```
moved {
  from = module.redis.aws_security_group.this
  to   = module.application_security_group.aws_security_group.redis
}
```

terraform/production/main.tf
```diff
@@ -203,11 +203,6 @@ moved {
   to   = module.application_roles.aws_iam_role_policy_attachment.task_execution_managed_policy
 }
 
-moved {
-  from = module.application.aws_iam_role_policy.parameter_access
-  to   = module.application_roles.aws_iam_role_policy.parameter_access
-}
-
```
Chirag-Bhatti, line 210: Seems to be a cyclic dependency when considered with the another moved block (on line 110) as below, which is the exact opposite 

```
moved {
  from = module.application_roles.aws_iam_role_policy.parameter_access
  to   = module.application.aws_iam_role_policy.parameter_access
}
```

The one on line 110 is correct so this can be removed

RachaelBooth: Did this move and get moved back again or something?

RachaelBooth (approved):

terraform/production/main.tf
Chirag-Bhatti: Yeah looks like it from the commit history. The moved blocks would have existed as and when needed in staging while being built progressively, but not in prod

PR #91, opened 2023-10-30
CLDC-2608: Deploy Production Updates

Chirag-Bhatti:
- Remove no longer required terraform moved blocks from Production
- Add a depends_on terraform statement to make apply commands smoother (in staging and prod)
- Update the docs about creating a new env from scratch
- Update the prod lock file for convenience

### DIFF
--- a/docs/environment_creation_from_scratch.md
+++ b/docs/environment_creation_from_scratch.md
@@ -1,24 +1,21 @@
 # Creating a new environment from scratch
 
-TODO: Update this once splitting out secrets and app roles is done, add instructions for sorting those and things to update in meta.
-
-When setting up a new environment;
+When setting up a new environment:
 1. DNS records will need to be created by DLUHC for certificate validation and to point at the cloudfront distribution and app load balancer
 1. Secret values will need to be filled in manually
-1. Need to set up deployments etc.
-1. Need to ensure anything that needs adding to the meta module (e.g. for roles access to ecr) is done, since some of that is currently manual
+1. The meta account will need updating with additional roles access to the ecr, since this is currently manual
+1. A new deployment pipeline will need to be setup
 
 ## The process
 
 ### Run an initial apply
 
-```terraform apply -target="module.networking" -target="module.front_door" -var="initial_create=true"```
+```terraform apply -target="module.networking" -target="module.front_door" -target="module.certificates" --target="module.application_roles" --target="module.application_secrets" var="initial_create=true"```
 
 (Need the networking explicitly otherwise we get an error creating the load balancer, there's some kind of dependency that terraform doesn't quite get)
 
-This will create the certificates, load balancer, cloudfront distribution, and some networking and other things.
-
-Potentially also do the secrets at this point (once they are pulled out of the application module) since they also have a manual step, so you can fill them out before the app itself is created and tries to read them.
+This will create the certificates, load balancer, cloudfront distribution, and some networking and other things needed for defining DNS records.
+It will also create some app roles and secrets which are necessary before creating the full app.
 
 ### Get DNS Records set up
 
@@ -30,6 +27,17 @@ This will be;
 * CNAME for the cloudfront domain pointing at the cloudfront distribution
 * CNAME for the load balancer domain pointing at the load balancer
 
+Once DLUHC has set up the records, check the certificates have been validated in AWS console.
+
+### Fill the application secrets
+
+When doing a full apply the complete application will be created, and will need to be able to read secrets. 
+Fill out the values for the secrets in AWS console before the full apply.
+
+### Update meta environment
+
+In the [meta/main.tf](../terraform/meta/main.tf) file, add the ARN of the task-execution role from the newly created environment to the ECR module's `allow_access_by_roles` parameter
+
 ### Run a full apply
 
 Once these have been done, run a complete apply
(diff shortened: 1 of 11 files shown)


terraform/production/main.tf
```diff
@@ -151,64 +116,10 @@ module "application" {
...
-
-moved {
-  from = module.application.aws_iam_role.task_execution
-  to   = module.application_roles.aws_iam_role.task_execution
-}
 
-moved {
-  from = module.application.aws_iam_role_policy_attachment.task_execution_managed_policy
-  to   = module.application_roles.aws_iam_role_policy_attachment.task_execution_managed_policy
+  depends_on = [module.database.rds_partial_connection_string_parameter_name]
```
Chirag-Bhatti, line 120: Terraform doesn't pick out this dependency of the application module, and therefore you can't apply until this resource exists (because a data block in the application module needs it)

terraform/staging/main.tf
```diff
@@ -121,6 +121,8 @@ module "application" {
   sentry_dsn_secret_arn                             = module.application_secrets.sentry_dsn_secret_arn
   sns_topic_arn                                     = module.monitoring.sns_topic_arn
   vpc_id                                            = module.networking.vpc_id
+
+  depends_on = [module.database.rds_partial_connection_string_parameter_name]
```
Chirag-Bhatti, line 125: See equivalent comment on prod file

terraform/production/main.tf
Chirag-Bhatti: This applied to staging as well

Chirag-Bhatti: For dev, we won't come across this as we create the "shared" infra first, followed by "per_review_app"

docs/environment_creation_from_scratch.md
Chirag-Bhatti, line 1: Not final or polished, but just adds a bit more to the steps now we've progressed the infra further

Chirag-Bhatti:
A general query - the secrets for Prod look to have all been set OK but for the `API_KEY`. I can't determine if it's necessary and if so, what it's value should be, as I can't see this is being used in PaaS Prod.

Do you have any more context about if this something to potentially strip from secrets and terraform for Prod?

docs/environment_creation_from_scratch.md
```diff
@@ -1,24 +1,21 @@
...
-1. Need to ensure anything that needs adding to the meta module (e.g. for roles access to ecr) is done, since some of that is currently manual
+1. The meta account will need updating with additional roles access to the ecr, since this is currently manual
+1. A new deployment pipeline will need to be setup
 
 ## The process
 
 ### Run an initial apply
 
-```terraform apply -target="module.networking" -target="module.front_door" -var="initial_create=true"```
+```terraform apply -target="module.networking" -target="module.front_door" -target="module.certificates" --target="module.application_roles" --target="module.application_secrets" var="initial_create=true"```
```
RachaelBooth, line 13: Do you need to specify the certificates module? I thought terraform automatically picked up that the certificates were a dependency of the front_door module and so created them.

terraform/production/main.tf
RachaelBooth: Huh, that's surprising.

Chirag-Bhatti: Yeah I'm surprised too. It could be because production was part way built before I bought it all up to date with the `main` branch. However I don't think having the `depends_on` causes confusion / does any harm, and it would be a lot of effort to prove it isn't necessary (I think we'd need to build another env from scratch to see for certain, but we've setup DNS for all of them now)

docs/environment_creation_from_scratch.md
Chirag-Bhatti: IIRC - when I was deploying the dev env for DNS (using `initial_create = true`), the `tf apply` would only create the load balancer certificate, even though the output from it's plan said it would create the cloudfront one as well. 

I reckon this is because of the line below in the `cloudfront.tf` file
```
acm_certificate_arn  = var.initial_create ? null : var.cloudfront_certificate_arn
```

So with `initial_create = true`, the certificate arn isn't actually needed for the various bits of the `front_door` module to be built, but the `front_door` module takes this as an input parameter, which is possibly why the plan is misleading.

Nonetheless, we still need both certificates to be made to ensure that we can give all DNS details to DLUHC, and adding this to the command was a way to ensure it 😀

RachaelBooth: I remember it working without when setting up the prod environment, because terraform considers both to be dependencies of the module even though neither is then actually used (as the load balancer one is only attached in a listener resource that is not created when initial_create is true). So I'm surprised if it didn't work for dev. 

Nonetheless the certificates module doesn't have other things in it that we'd want not to create yet, and I'm assuming copy-paste rather than typing, so 🤷‍♀️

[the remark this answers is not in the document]
Chirag-Bhatti: I think this was an unnecessary change by me - terraform should pick up the dependencies of the `application_secrets` module (it needs a part of the `application_roles` module) and create them as necessary, rather than us commanding it create the whole module. I've updated the command to remove this 😀

Chirag-Bhatti: Hmm very odd. I'm unsure if it's because of some of the refactoring work we did for `[CLDC-2811](https://dluhcdigital.atlassian.net/browse/CLDC-2811)` that made me see this on dev. 

I've removed `-target=module.certificates` from the command for now, and will add an instruction to try this target if the certificates haven't been made as expected as part of [CLDC-2763](https://dluhcdigital.atlassian.net/browse/CLDC-2763)

SamSeed-Softwire (commented):
I have nothing to add!

Chirag-Bhatti:
> A general query - the secrets for Prod look to have all been set OK but for the `API_KEY`. I can't determine if it's necessary and if so, what it's value should be, as I can't see this is being used in PaaS Prod.
> 
> Do you have any more context about if this something to potentially strip from secrets and terraform for Prod?

Some more info about this for future context:

- There is this Slack thread [here](https://softwire.slack.com/archives/C05BDC4JBRN/p1698685815122829) about the `API_KEY` and authenticating API requests
- There is another Slack thread [here](https://softwire.slack.com/archives/C03QXAE3Y3W/p1698920890683819), where we successfully made API requests to the app and explored the authentication behaviour
- We concluded we should remove any vars to do with the API in this PR (for all our envs in terraform), because the API is not being used right now

RachaelBooth (approved):

PR #92, opened 2023-10-31
CLDC-2921: SNS email subscriptions

SamSeed-Softwire:
I've tested this and it works. The team DL email address is saved in secrets manager, and I've just updated the team DL to include Rachael, Chirag and myself (this can easily be updated via HYP of course).

I've added some docs [here](https://dluhcdigital.atlassian.net/l/cp/wcf0bRKp) about the fact that when applying a new subscription in an environment, you'll get an error if you haven't already created the secret using `terraform apply -target="module.monitoring" -var="create_secrets_first=true"` and also set its value. There is a more comprehensive guide on this in `docs/environment_creation_from_scratch.md`.

### DIFF
--- a/terraform/staging/variables.tf
+++ b/terraform/staging/variables.tf
@@ -1,3 +1,9 @@
+variable "create_secrets_first" {
+  type        = bool
+  description = "Setting to true will avoid creating any infrastructure for which a terraform apply would fail if certain secret values are not set"
+  default     = false
+}
+
 variable "initial_create" {
   type        = bool
   description = "Set to true during an initial apply in a new environment, to avoid cloudfront referencing certificates before they've been validated manually"
--- a/terraform/staging/variables.tf
+++ b/terraform/staging/variables.tf
@@ -1,9 +1,3 @@
-variable "create_secrets_first" {
-  type        = bool
-  description = "Setting to true will avoid creating any infrastructure for which a terraform apply would fail if certain secret values are not set"
-  default     = false
-}
-
 variable "initial_create" {
   type        = bool
   description = "Set to true during an initial apply in a new environment, to avoid cloudfront referencing certificates before they've been validated manually"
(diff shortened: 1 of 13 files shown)


RachaelBooth:
> I've added some docs [here](https://dluhcdigital.atlassian.net/l/cp/wcf0bRKp) about the fact that when applying a new subscription in an environment, you'll likely get an error because the secret will get created, but won't have its value set yet, and then Terraform will try to use that secret (which has no value) to create the subscription.

It seems like it would be worth having an option to create the secret without terraform trying to read its value to avoid this problem (a bit like the initial_create stuff works) - although it might well not be worth the effort. 

In either case, does anything needing adding to the doc about creating an environment from scratch (in the docs folder in the repo)?

SamSeed-Softwire:
> > I've added some docs [here](https://dluhcdigital.atlassian.net/l/cp/wcf0bRKp) about the fact that when applying a new subscription in an environment, you'll likely get an error because the secret will get created, but won't have its value set yet, and then Terraform will try to use that secret (which has no value) to create the subscription.
> 
> It seems like it would be worth having an option to create the secret without terraform trying to read its value to avoid this problem (a bit like the initial_create stuff works) - although it might well not be worth the effort.
> 
> In either case, does anything needing adding to the doc about creating an environment from scratch (in the docs folder in the repo)?

I've implemented this suggestion and have updated the docs 👍

terraform/staging/variables.tf
```diff
@@ -1,5 +1,11 @@
+variable "create_secrets_first" {
```
SamSeed-Softwire, line 1: I could potentially have used `initial_create` here, but felt it was cleaner to keep that focused solely on the stuff needed for DNS setup etc.

RachaelBooth (approved):
If you end up with a spare bit of time, it may be worth linking this up (at least doc-wise) with the creating/filling in application secrets steps more - but it's not worth blocking the main bit of this work on, and there are probably more important things remaining.

PR #94, opened 2023-11-06
Front door bucket names

RachaelBooth:
Changes needed to deploy shared infra to the dev environment.


terraform/modules/front_door/cloudfront_access_log_bucket.tf
```diff
@@ -6,7 +6,7 @@ resource "aws_s3_bucket" "cloudfront_access_logs" {
   #checkov:skip=CKV_AWS_18: access log bucket doesn't need access logging itself
   #checkov:skip=CKV_AWS_145: access log buckets not compatible with kms encryption, see https://docs.aws.amazon.com/AmazonS3/latest/userguide/troubleshooting-server-access-logging.html
   #checkov:skip=CKV_AWS_144: cross region replication not required for access logs
-  bucket = "${var.prefix}-cloudfront-logs"
+  bucket = local.bucket_names.cloudfront_logs
```
RachaelBooth, line 9: It made sense to me to set this up as well since it's in the same module, even though we're only using it for the load balancer logs bucket.

terraform/modules/front_door/main.tf
```diff
@@ -13,3 +13,12 @@ terraform {
     }
   }
 }
+
+data "aws_caller_identity" "current" {}
```
RachaelBooth, line 17: Could have chosen to use something random instead, or let us define a string in each case, but this seemed to make sense and is highly likely to be unique.

SamSeed-Softwire (approved):
Lgtm!

PR #96, opened 2023-11-06
CLDC-2763: Update create env from scratch document instructions

Chirag-Bhatti:
As per title

I've been staring at this a little while (last week and for this ticket), so I wanted to get your thoughts on whether you thought it had all the steps and the right level of detail before merging 😀


docs/environment_creation_from_scratch.md
```diff
@@ -1,54 +1,112 @@
...
 
 ## The process
 
+### Create a new terraform folder
+
+If the environment hasn't been created / defined before:
+
+In the terraform folder, add a new sub-folder with the name of the environment.
+
+Add a `main.tf` file to the folder, and define any infrastructure here (e.g. modules to use). You may also add
```
SamSeed-Softwire, line 20: You could maybe also call out any local vars that need to be set correctly e.g. `provider_role_arn` (and maybe `app_host` and `load_balancer_domain_name`). Also probs fine without.

docs/environment_creation_from_scratch.md
```diff
@@ -1,54 +1,112 @@
...
+the minimum required for defining the DNS records that DLUHC will need to add. It will also create some app roles and 
+secrets; which are necessary before creating the full app).
 
 ### Get DNS Records set up
 
-Ask DLUHC to set up DNS records required.
+Ask DLUHC to add the DNS records for the new environment.
 
-This will be;
+There will be four records, which will be;
```
SamSeed-Softwire, line 47: You could maybe add an example of each record here to make things super clear? Not necessary though.

docs/environment_creation_from_scratch.md
```diff
@@ -1,54 +1,112 @@
...
+Alternatively, if the db migration stage ran successfully, you can run a `db:seed` rake task separately (also using the 
+`ad_hoc` task definition), to populate the DB with some initial data.
 
-### Set up the deployment
+### Run a complete deployment
 
-In the Core App codebase [here](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data), you will need create a new pipeline for the environment (you can re-use jobs and configuration in existing pipelines where appropriate)
-For the `aws_deploy` workflow, ensure you pass in the new account id, resource prefix and environment name
\ No newline at end of file
+Finally, run the deployment pipeline completely to bring the infrastructure and application to a complete and ready-to-use
```
SamSeed-Softwire, line 111: How does this differ from running the deployment as mentioned in the `### Set up the deployment` section above?

Chirag-Bhatti: You need to run the deployment first to update the task definitions to pull the correct container image i.e. pull an image with a name/tag that exists (rather than `latest` which doesn't). However if there is no DB setup, the deployment will fill on the migrations step (which is OK). 

You then have to go and make the DB (which you can only do if the task definitions are up to date, so you can run an ad-hoc task to do this which will pull the right image 😅). Then you'll be able to re-run the deployment from start to end successfully. 

Hope that makes sense 😀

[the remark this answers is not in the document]
Chirag-Bhatti: See my comment above on Sam's question 😀

docs/environment_creation_from_scratch.md
Chirag-Bhatti: Added 👍

docs/environment_creation_from_scratch.md
Chirag-Bhatti: Added 👍

docs/add_dns_records_dluhc_request_example.png
Chirag-Bhatti, line 1: N.B. this example doesn't contain real values, just dummy ones that look like similar scramble

docs/environment_creation_from_scratch.md
Chirag-Bhatti: I've updated some of the instruction wording to make the above clearer 😃

SamSeed-Softwire: Nice one, thanks!

SamSeed-Softwire (approved):
Lgtm :)

docs/environment_creation_from_scratch.md
```diff
@@ -100,13 +107,13 @@ Once setup, run the deployment pipeline in order to update the ECS task definiti
 ### Set up the database
 
 If you're not restoring the database from a backup / migrating it from elsewhere, and the db migration stage failed
-above, run a `db:setup` rake task (using the `ad_hoc` task definition to spin it up from the app image, with a command 
-override).
+in the deployment above, run a `db:setup` rake task (using the `ad_hoc` task definition to spin it up from the app image, 
+with a command override).
```
RachaelBooth, line 111: Is your editor doing automatic line breaks or something? (It's not a problem when considering what it looks like as markdown, but it's adding extra diffs)

RachaelBooth (approved):

PR #97, opened 2023-11-07
CLDC-2535: Set up action to deploy infra for review app

RachaelBooth:


.github/workflows/create_review_app_infra.yml
```diff
@@ -0,0 +1,47 @@
...
+      id-token: write
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Set up Terraform
+        uses: hashicorp/setup-terraform@v2
+        with:
+          terraform_version: 1.5.2
```
RachaelBooth, line 32: We should look at upgrading this version at some point, but I'm matching that used in the checks for now.

terraform/development/per_review_app/main.tf
```diff
@@ -14,7 +14,6 @@ terraform {
     encrypt        = true
     key            = "core-development-per-review-app.tfstate"
     region         = "eu-west-2"
-    role_arn       = "arn:aws:iam::815624722760:role/developer"
```
RachaelBooth, line 17: The assumption here is that terraform will be run from a role that natively has access to the necessary items for the state (rather than defining a new role, in part because terraform is annoying about role chaining).

RachaelBooth: N.B. As developers, we can do this by running terraform while in the developer role in the meta account.

terraform/development/shared/main.tf
```diff
@@ -153,6 +153,16 @@ module "database" {
   vpc_id                      = module.networking.vpc_id
 }
 
+module "deployment_role" {
```
RachaelBooth, line 156: This might at some point extend to be used for more automated terraform deployments, but for now it's only set up for allowing the review app ones.

terraform/modules/github_actions_access/app_repo_policies.tf
```diff
@@ -0,0 +1,31 @@
+data "aws_iam_policy_document" "push_images" {
```
RachaelBooth, line 1: Arguably could refactor to put this with the ecr definition, but it was already in this module so that seems a bit more effort than it's worth right now.

terraform/modules/github_actions_access/repo_roles.tf
```diff
@@ -15,47 +24,27 @@ data "aws_iam_policy_document" "app_repo_assume_role" {
...
+      values   = ["repo:${each.value}:*"]
       variable = "token.actions.githubusercontent.com:sub"
     }
   }
 }
 
-resource "aws_iam_role" "app_repo" {
-  name               = "core-application-repo"
-  assume_role_policy = data.aws_iam_policy_document.app_repo_assume_role.json
+moved {
```
RachaelBooth, line 33: TODO: Remove the moved blocks on merge.

terraform/development/shared/main.tf
```diff
@@ -153,6 +153,16 @@ module "database" {
   vpc_id                      = module.networking.vpc_id
 }
 
+module "deployment_role" {
+  source = "../../modules/terraform_deployment"
+
+  prefix = local.prefix
+  assume_from_role_arns = [
+    "arn:aws:iam::815624722760:role/developer",
+    "arn:aws:iam::815624722760:role/core-infrastructure-repo"
```
RachaelBooth, line 162: We've got a future ticket already to get some outputs from meta - leaving making this not hardcoded to be part of that.

.github/workflows/destroy_review_app_infra.yml
```diff
@@ -0,0 +1,49 @@
...
+          aws-region: ${{ env.aws_region }}
+          role-to-assume: ${{ env.app_repo_role }}
+
+      - name: Terraform destroy
+        working-directory: terraform/development/per_review_app
+        env:
+          workspace: ${{ inputs.key }}
+        run: |
+          terraform init
+          terraform workspace select -or-create $workspace
```
RachaelBooth, line 46: This is still `-or-create` in case it ends up running twice for any reason - would be nice to not break if the workspace has already been deleted.

terraform/meta/main.tf
```diff
@@ -122,7 +122,38 @@ data "aws_caller_identity" "current" {}
...
+  meta_account_id = data.aws_caller_identity.current.account_id
+  repositories = {
+    application = {
+      name = "communitiesuk/submit-social-housing-lettings-and-sales-data",
+      policies = [
+        { key = "push_ecr_images", arn = module.ecr.push_images_policy_arn },
+        { key = "access_non_prod_state", arn = module.non_prod_backend.state_access_policy_arn }
+      ]
+    },
+    infrastructure = {
```
RachaelBooth, line 134: N.B. Right now this isn't actually necessary, but I'd got so close before the move to have the app repo run this bit that it seemed worth leaving in, given we're expecting to set up automated terraform deployments at least for the dev environment (hopefully before too long!)

terraform/meta/main.tf
```diff
@@ -122,7 +122,38 @@ data "aws_caller_identity" "current" {}
...
+      name     = "communitiesuk/submit-social-housing-lettings-and-sales-data-infrastructure"
+      policies = []
+    }
+  }
+}
+
+moved {
```
RachaelBooth, line 141: N.B. These moved blocks are left in for the review for clarity, but I'll remove them afterwards (they are useful for seeing what's going on, but aren't needed once the terraform has been run)

kosiakkatrina (approved):
Looks good, a few files are missing new lines at the end tho!
Would it make sense to update readme as part of this?

PR #102, opened 2023-11-10
CLDC-none: enable AWS Shield for Prod

Chirag-Bhatti:
- Enable flag for aws shield on prod
- Remove unrequired aws shield application layer auto response for the load balancer. To have, shield requires you have a web ACL associated with the LB, with a rate limit rule also applied. We don't have a web ACL for it (or use an existing one such as that associated with the CloudFront distribution) as we decided to not have WAF on the LB (only Cloudfront). See comments on PR [here](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data-infrastructure/pull/51)


RachaelBooth (approved):
I think you may need to manually set things up in the AWS console to allow use of shield as well.

PR #108, opened 2023-11-30
Add alarms for max CPU and memory to catch single task problems

RachaelBooth:
This adds alarms for each ecs cluster that look at the max CPU / memory utilisation instead of the average. This would catch issues like the sidekiq one we saw the other day, where one task is running at full throttle for a long period of time but the other is normal, and thus the average is still quite low. 

I think both types are still valuable - the new ones will only alarm if max values are consistently high for 30 minutes, while the old will let us know that the average is high on a much shorter time basis. The old ones imply a potential overall capacity problem, the new ones something potentially wrong with one task in the system.

### DIFF
--- a/terraform/modules/application/alarms.tf
+++ b/terraform/modules/application/alarms.tf
@@ -1,6 +1,6 @@
-resource "aws_cloudwatch_metric_alarm" "app_cpu" {
+resource "aws_cloudwatch_metric_alarm" "app_average_cpu" {
   alarm_actions             = [var.sns_topic_arn]
-  alarm_name                = "${var.prefix}-app-cpu"
+  alarm_name                = "${var.prefix}-app-average-cpu"
   comparison_operator       = "GreaterThanOrEqualToThreshold"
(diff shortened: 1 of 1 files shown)


terraform/modules/application/alarms.tf
```diff
@@ -1,6 +1,6 @@
-resource "aws_cloudwatch_metric_alarm" "app_cpu" {
+resource "aws_cloudwatch_metric_alarm" "app_average_cpu" {
   alarm_actions             = [var.sns_topic_arn]
-  alarm_name                = "${var.prefix}-app-cpu"
+  alarm_name                = "${var.prefix}-app-average-cpu"
```
RachaelBooth, line 3: I think the name change is wanted, which forces a new resource anyway hence the lack of moved blocks.

RachaelBooth:
I'll update the alerts docs in confluence after this is done.

kosiakkatrina (approved):

PR #112, opened 2023-12-21
Suppress missing data in alarms for review apps

RachaelBooth:
We keep getting emails for these alarms going from insufficient data -> ok shortly after a review app is first set up.

For review apps, this is annoying and there is little risk by potentially ignoring lack of data.
For staging and prod, missing data should not happen because the environments are already set up (data shouldn't be missing during a release, just when the infra setup happens) so is concerning, hence we'll have these metrics alarm for missing data.


RachaelBooth:
N.B. I've not applied this anywhere yet - the key test is review apps and the code needs to be on main for that.

kosiakkatrina (approved):

PR #115, opened 2024-01-05
Cldc 3115 reduce dev db spec

RachaelBooth:
This looks like it will meet our actual usage needs very comfortably - it does assume we are not using too many review apps heavily at once, which feels like a safe assumption.

We may even be able to get away with micro instead of small, but the savings there are much smaller scale so it feels best to start with this.

### DIFF
--- a/terraform/modules/rds/variables.tf
+++ b/terraform/modules/rds/variables.tf
@@ -13,6 +13,12 @@ variable "backup_retention_period" {
   description = "The number of days to retain db backups for. If 0 then the database will not be backed up"
 }
 
+variable "create_replica" {
+  type        = bool
+  description = "If true, creates a replica db"
+  default     = false
+}
+
 variable "database_port" {
   type        = number
   description = "The network port the database runs on"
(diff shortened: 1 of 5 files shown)


RachaelBooth:
N.B. There will be downtime (for all review apps) when applying this change to the dev db.

terraform/modules/rds/variables.tf
```diff
@@ -30,7 +36,8 @@ variable "enable_primary_deletion_protection" {
 
 variable "enable_replica_deletion_protection" {
   type        = bool
-  description = "Whether the replica database should have deletion protection enabled"
+  description = "Whether the replica database (if create_replica is true) should have deletion protection enabled"
+  default     = true
```
kosiakkatrina, line 40: What is the benefit of setting the default here?

RachaelBooth: My reasoning here was it meant not needing to specify a value when there is no replica (which is a weird thing to have to do) - annoyingly you can't get interesting conditions that reference other variables, but true seems like a reasonable default here if you do have a replica and don't specify.

kosiakkatrina: Yup, that makes sense

kosiakkatrina (approved):

PR #122, opened 2024-06-17
CLDC-3481: Add alarm for cloudfront 5xx error rate

RachaelBooth:


RachaelBooth:
This has less change than it looks!

Annoyingly because of how cloudfront works the new alarm needs to be in us-east-1, and so the topic does too. I've pulled the monitoring module apart a bit because that was what ended up making most sense for sorting out multiple regions of topics.

terraform/development/shared/main.tf
```diff
@@ -187,15 +188,51 @@ module "front_door" {
...
-  create_email_subscription = false
+  prefix = local.prefix
+
+  initial_create = var.initial_create
+}
+
+module "monitoring_topic_main" {
+  source = "../../modules/monitoring_topic"
+
+  create_email_subscription = true
```
RachaelBooth, line 202: We actually do have a subscription already, but it's not managed in terraform. I think it should be brought in.

terraform/modules/monitoring_secrets/kms.tf
```diff
@@ -1,27 +1,19 @@
 resource "aws_kms_key" "this" {
-  count = var.create_email_subscription ? 1 : 0
```
RachaelBooth, line 2: Everything in the module ended up with this condition, so if not wanted the entrypoint should just not call the module.

terraform/staging/main.tf
```diff
@@ -295,15 +296,36 @@ module "networking" {
   vpc_flow_cloudwatch_log_expiration_days = 60
 }
 
-module "monitoring" {
-  source = "../modules/monitoring"
+module "monitoring_secrets" {
+  source = "../modules/monitoring_secrets"
 
-  create_email_subscription = false
```
RachaelBooth, line 301: Again, we did already actually have one just not in terraform.

kosiakkatrina (approved):

PR #127, opened 2024-07-23
CLDC-2616: Add scheduled scaling

RachaelBooth:
* Stops review app services and database outside of working hours
* Scales down number of app instances for staging and production outside of (more generous) working hours when they get little usage


terraform/staging/main.tf
```diff
@@ -87,6 +87,18 @@ module "application" {
   sidekiq_task_desired_count = 2
   sidekiq_task_memory        = 8192
 
+  out_of_hours_scale_down = {
+    enabled = true
+    timings = {
+      workday_start = "0 7"
+      workday_end   = "30 19"
+    }
```
RachaelBooth, line 95: I've looked at the load balancer graphs for number of requests to find sensible times here (erring on the side of caution).

Note in this case we're only scaling down the number of app instances to 2 rather than 4 - the application will still function, just aiming to handle fewer requests.

(We should consider autoscaling at some point, but that's not in scope right now)

terraform/development/shared/main.tf
```diff
@@ -143,6 +143,15 @@ module "database" {
   skip_final_snapshot                = true
   instance_class                     = "db.t4g.small"
 
+  scheduled_stop = {
+    enabled = true
+    timings = {
+      workday_start = "30 7"
+      workday_end   = "0 20"
+    }
```
RachaelBooth, line 151: These timings are a bit wider than the app to give the db time to start and to have a maintenance window.

terraform/production/main.tf
```diff
@@ -87,6 +87,18 @@ module "application" {
...
 
+  out_of_hours_scale_down = {
+    enabled = true
+    timings = {
+      workday_start = "0 7"
+      workday_end   = "30 19"
+    }
+    scale_to = {
+      app     = 2
+      sidekiq = 2
```
kosiakkatrina, line 98: Could there be a situation where something long is running on sidekiq? I guess it would just fail and restart on another instance, but is there a way to avoid shutting down busy instances (assuming this is not doing that already)?

kosiakkatrina (approved):

terraform/production/main.tf
RachaelBooth: Sidekiq is already set to 2 for prod, so this doesn't actually change it (otherwise you're right I'd be a bit worried we'd lose ongoing work for a long running task even though it would be restarted on another instance - should probably find a way to make that less of a problem but I suspect that may be a bit tricky to do)

PR #128, opened 2024-07-25
Budget alerts

RachaelBooth:


terraform/development/shared/main.tf
```diff
@@ -68,6 +68,15 @@ locals {
   redis_port       = 6379
 }
 
+module "budget" {
+  source = "../../modules/budget"
+
+  cost_limit = 800
```
RachaelBooth, line 74: Numbers are based on current spend with some vague calculation about what we expect recent changes to save - I expect to check them again in a few months (particularly this one because the dev account usage is the most variable).

RachaelBooth:
Note to self: need to import the existing budgets when releasing this (done for staging).

kosiakkatrina (approved):

PR #129, opened 2024-07-26
CLDC-2616: Adjust alarms based on out of hours scaling

RachaelBooth:


terraform/modules/application/alarms.tf
```diff
@@ -1,3 +1,22 @@
+locals {
+  app_scales_to_zero               = var.out_of_hours_scale_down.enabled ? var.out_of_hours_scale_down.scale_to.app == 0 : false
+  sidekiq_scales_to_zero           = var.out_of_hours_scale_down.enabled ? var.out_of_hours_scale_down.scale_to.sidekiq == 0 : false
+  app_desired_count_for_alarms     = (!var.out_of_hours_scale_down.enabled || local.app_scales_to_zero) ? var.app_task_desired_count : var.out_of_hours_scale_down.scale_to.app
+  sidekiq_desired_count_for_alarms = (!var.out_of_hours_scale_down.enabled || local.sidekiq_scales_to_zero) ? var.sidekiq_task_desired_count : var.out_of_hours_scale_down.scale_to.sidekiq
+}
+
+check "suppress_missing_alarm_data_when_scaling_to_zero" {
```
RachaelBooth, line 8: This doesn't block you from continuing when these fail, but will show a warning.

terraform/modules/application/alarms.tf
```diff
@@ -278,7 +298,7 @@ resource "aws_cloudwatch_metric_alarm" "healthy_hosts_count" {
   ok_actions                = var.suppress_ok_notifications ? [] : [var.sns_topic_arn]
   period                    = 60
   statistic                 = "Minimum"
-  threshold                 = var.app_task_desired_count
+  threshold                 = local.app_desired_count_for_alarms
```
RachaelBooth, line 301: This does make this alarm less stringent - given that we almost inevitably have the unhealthy hosts count alarm as well, I think we're ok with that.

kosiakkatrina (approved):

PR #131, opened 2024-08-14
Rename all references of dluhc to mhclg excl urls

Dinssa:
This PR changes "dluhc" text to "mhclg".

A few services are named "dluhc", without first changing those this PR will be premature to merge in.

I've also excluded web urls, until the domain is changed from levellingup.gov.uk.


db-migration/run.sh
```diff
@@ -2,7 +2,7 @@
 
 echo 'Logging into Gov PaaS'
 
-if cf login -a api.london.cloud.service.gov.uk -o "dluhc-core" -u ${CF_USERNAME} -p ${CF_PASSWORD} -s ${CF_SPACE}; then
```
RachaelBooth, line 5: This refers to a specific space in gov PaaS that was called dluhc-core, which no longer exists as PaaS was decommissioned. Given that, it seems odd to me to change this (since it wouldn't work) - I wonder if we might be better to remove the db-migration code entirely at this point? (If so, the s3 migration code could go with the same reasoning)

PR #132, opened 2024-08-22
CLDC-3616: Create certificates for new url

RachaelBooth:


terraform/development/shared/main.tf
```diff
@@ -59,6 +59,9 @@ locals {
   app_host                  = "review.submit-social-housing-data.levellingup.gov.uk"
   load_balancer_domain_name = "review.lb.submit-social-housing-data.levellingup.gov.uk"
 
+  new_app_host                  = "review.submit-social-housing-data.communities.gov.uk"
```
RachaelBooth, line 62: Naming here is very temporary - we can't get rid of the old certificates yet, but the plan is once we are switched over and have a redirect in place we can remove the old ones and this code will look as it did before (just with different text for app_host and load_balancer_domain_name).

terraform/modules/certificates/certificates.tf
```diff
@@ -2,6 +2,8 @@ resource "aws_acm_certificate" "cloudfront" {
   domain_name       = var.cloudfront_domain_name
   validation_method = "DNS"
 
+  subject_alternative_names = var.cloudfront_additional_names
```
RachaelBooth, line 5: Cloudfront only allows one certificate attached at once, so will need a cert for both in order to do the switchover.

You can attach multiple certificates to one load balancer though.

kosiakkatrina (approved):

PR #136, opened 2024-09-04
CLDC-3616: Attach new certificates to cloudfront and load balancer

RachaelBooth:
This (combined with the DNS changes) allows the site to be reached via either the old or new url


RachaelBooth:
Again, a lot of this will go away after the switch.

PR #137, opened 2024-09-04
CLDC-3616: Swap url used by cloudfront for load balancer

RachaelBooth:


RachaelBooth:
This should be applied after attaching the new certificates

kosiakkatrina (approved):

PR #138, opened 2024-09-04
CLDC-3616: Update app host env variable

RachaelBooth:


RachaelBooth:
This means that links generated by the app (e.g. sent out in emails) will use the new url - will want to be coordinated with comms etc.

kosiakkatrina (approved):

PR #140, opened 2024-09-10
Add user credentials for performance tests

kosiakkatrina:


terraform/production/main.tf
```diff
@@ -140,6 +140,8 @@ module "application" {
   redis_connection_string                           = module.redis.redis_connection_string
   sentry_dsn_secret_arn                             = module.application_secrets.sentry_dsn_secret_arn
   sns_topic_arn                                     = module.monitoring_topic_main.sns_topic_arn
+  staging_performance_test_email_secret_arn         = module.application_secrets.staging_performance_test_email_secret_arn
+  staging_performance_test_password_secret_arn      = module.application_secrets.staging_performance_test_password_secret_arn
```
RachaelBooth, line 144: It seems a bit odd to me that all the environments need this, is there a nice way to just only sometimes provide it?

RachaelBooth (approved):

terraform/production/main.tf
kosiakkatrina: Ah, yes, I set the default arn to null, would that work?

PR #143, opened 2024-09-30
CLDC-3516: Add collection resources bucket

Dinssa:


RachaelBooth (approved):

terraform/modules/backend/.terraform.lock.hcl
```diff
@@ -2,64 +2,64 @@
 # Manual edits may be lost in future updates.
 
 provider "registry.terraform.io/hashicorp/aws" {
-  version     = "5.9.0"
+  version     = "5.69.0"
```
RachaelBooth, line 5: We could get rid of this from this folder instead of updating - there shouldn't be any reason to run terraform from it.

terraform/modules/collection_resources/bucket.tf
```diff
@@ -0,0 +1,120 @@
...
+  block_public_policy     = false
+  ignore_public_acls      = false
+  restrict_public_buckets = false
+}
+
+resource "aws_s3_bucket_policy" "public_read" {
+  bucket = aws_s3_bucket.collection_resources.id
+
+  #checkov:skip=CKV_AWS_70: Public access block is intentionally disabled for this bucket
+  policy = jsonencode({
```
RachaelBooth, line 38: Could this come out into a policy document data resource, rather than json encoding an object? (and similar elsewhere) 
We're a bit inconsistent across this codebase, but generally aiming towards that method.

(I know this isn't new since my last review, but would be a good bit of tidying)

RachaelBooth (approved):

PR #145, opened 2024-09-30
CLDC-3616: Create certificates for just the new domain

RachaelBooth:


terraform/development/shared/main.tf
```diff
@@ -153,10 +153,10 @@ module "certs_for_new_domain" {
...
   }
 
-  cloudfront_domain_name         = local.new_app_host
-  cloudfront_additional_names    = [local.app_host]
-  load_balancer_domain_name      = local.new_load_balancer_domain_name
-  load_balancer_additional_names = [local.load_balancer_domain_name]
+  cloudfront_domain_name         = local.app_host
+  cloudfront_additional_names    = [local.old_app_host]
+  load_balancer_domain_name      = local.load_balancer_domain_name
+  load_balancer_additional_names = [local.old_load_balancer_domain_name]
```
RachaelBooth, line 159: Because of the change above, this stays the same and the `certificates` module above will remove the old domain certs it previously managed and create ones for the new domain only (instead of the dual certs covering both).

kosiakkatrina (approved):

PR #152, opened 2024-11-20
CLDC-3695 Update docs

kosiakkatrina:


terraform/modules/front_door/ip_list.tf
```diff
@@ -8,7 +8,7 @@ locals {
     "87.224.105.250/32",
     "87.224.116.242/32",
     "45.150.142.210/32",
-    # DLUHC
+    # MHCLG
```
RachaelBooth, line 11: Should check if we still need this list at some point - didn't we switch it to geo-restriction?

But not necessarily now, it's just reminded me.

RachaelBooth (approved):

terraform/modules/front_door/ip_list.tf
kosiakkatrina: Yeah, we did. I think at the time it wouldn't apply without this, I can't remember what exactly the reason was

PR #154, opened 2024-12-16
CLDC-3810: Automatically redeploy prod and staging when collection year rolls over

RachaelBooth:


terraform/modules/application/scheduled_redeploy.tf
```diff
@@ -0,0 +1,100 @@
+resource "aws_scheduler_schedule" "app_collection_rollover_redeploy" {
+  #checkov:skip=CKV_AWS_297: CMK not required here and seems to cause issues
+  count = var.collection_rollover_redeploy_enabled ? 1 : 0
+
+  name = "${var.prefix}-app-collection-rollover-redeploy"
+
+  schedule_expression          = "cron(0 0 1 4 ? *)"
```
kosiakkatrina, line 7: Sometimes things on AWS are an hour off, like logs, or the time at which things are uploaded to S3. I'm not sure if it would impact this at all, but would it make sense to redeploy at 1am ?

kosiakkatrina (approved):

terraform/modules/application/scheduled_redeploy.tf
RachaelBooth: This should be set to execute in UTC (so by April it will be 1am local time), which means it will be fine - AWS times on logs / file uploads I think are always in UTC, the important one for this will be what day the app thinks it is which is probably based on UTC (but if not will still be ok this way round).

PR #157, opened 2025-01-06
CLDC-3819: Create lambda to send monitoring alerts to slack

RachaelBooth:


terraform/modules/slack_alert_lambda/code/slack_alerts.rb
```diff
@@ -0,0 +1,324 @@
...
+end
+
+def alarm_activated_message(alarm)
+  {
+    blocks: [
+      { 
+        type: 'header',
+        text: {
+          type: 'plain_text',
+          text: "#{alarming_slack_emoji} #{ENV['ENVIRONMENT']} alarm triggered: #{alarm[:name]}"
```
kosiakkatrina, line 69: Is AlarmName something like `core-prod-cloudfront-5xx-error-rate` ? I'm on the fence on whether it's pleasant enough to see in a header. Would it be a pain to map it to something more human?

kosiakkatrina (approved):

terraform/modules/slack_alert_lambda/code/slack_alerts.rb
RachaelBooth: It is - and I think we need it in there somewhere to help find things on AWS, although it doesn't have to be the header. We could probably map it, or maybe use some string manipulation to get e.g. "Cloudfront 5xx Error Rate"?

Might depend how much forms work we have ready!

PR #172, opened 2026-03-05
Upgrade Postgres

oscar-richardson-softwire:
Part of https://mhclgdigital.atlassian.net/browse/CLDC-4217.


samyou-softwire (approved):
LGTM, one small comment suggestion

do we need to run a deployment after merging this?

oscar-richardson-softwire:
> LGTM, one small comment suggestion
> 
> do we need to run a deployment after merging this?

Thanks for the suggestion, actioned.

I don't think we need to run a deployment as Terraform won't detect any changes anyway, due to `ignore_changes = [engine_version]`

samyou-softwire (approved):

PR #175, opened 2026-04-22
CLDC-4127: Re-enable meta deployment

natdeanlewissoftwire:
Closes [CLDC-4127](https://mhclgdigital.atlassian.net/browse/CLDC-4127).

Re-enables the meta (integration) manual deployments, and updates/codifies a couple of pre-existing related policies etc. to allow this. I've done this without changing to the existing staging/prod setup, and have used the approach there as a template for safety. All changes are purely meta-based and additive/role remapping rather than destructive.

I've tested this on branch but will also confirm the manual workflow runs seamlessly on main once merged (though there won't be any changes to apply ofc!) 
UPDATE - confirmed ✅ https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data-infrastructure/actions/runs/24829672634

### DIFF
--- a/terraform/meta/.terraform.lock.hcl
+++ b/terraform/meta/.terraform.lock.hcl
@@ -1,6 +1,26 @@
 # This file is maintained automatically by "terraform init".
 # Manual edits may be lost in future updates.
 
+provider "registry.terraform.io/hashicorp/archive" {
+  version     = "2.7.1"
+  constraints = "~> 2.7.0"
+  hashes = [
+    "h1:A7EnRBVm4h9ryO9LwxYnKr4fy7ExPMwD5a1DsY7m1Y0=",
+    "zh:19881bb356a4a656a865f48aee70c0b8a03c35951b7799b6113883f67f196e8e",
+    "zh:2fcfbf6318dd514863268b09bbe19bfc958339c636bcbcc3664b45f2b8bf5cc6",
+    "zh:3323ab9a504ce0a115c28e64d0739369fe85151291a2ce480d51ccbb0c381ac5",
+    "zh:362674746fb3da3ab9bd4e70c75a3cdd9801a6cf258991102e2c46669cf68e19",
+    "zh:7140a46d748fdd12212161445c46bbbf30a3f4586c6ac97dd497f0c2565fe949",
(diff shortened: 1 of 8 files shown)


terraform/meta/main.tf
```diff
@@ -54,8 +62,8 @@ module "budget" {
 }
 
 # We create two backends for managing the terraform state of different accounts:
-# non_prod manages meta, development and staging
-# prod manages just production
+# non_prod manages development and staging
+# prod manages production and meta
```
natdeanlewissoftwire, line 66: I'm not sure if this is intentional, but the docs should reflect reality. Do we want to change this? Should be a separate ticket if so imo

terraform/modules/github_actions_access/repo_roles.tf
```diff
@@ -17,7 +17,7 @@ data "aws_iam_policy_document" "repo_assume_role" {
 
     condition {
       test     = "StringLike"
-      values   = ["repo:${each.value.name}:*"]
+      values   = [for repo in concat([each.value.name], each.value.additional_trusted_repo_names) : "repo:${repo}:*"]
```
natdeanlewissoftwire, line 20: This expansion of the repo trust to include additional repos (in particular expanding app to also trust infra) was previously manually added to AWS and is required for the meta manual deploy to run. Adding this just codifies what was already in AWS

terraform/meta/.terraform.lock.hcl
samyou-softwire, line 1: should this have updated? don't think we've changed any providers

terraform/meta/main.tf
samyou-softwire: feels like it'd be more hassle than worth to try and migrate this now

samyou-softwire (approved):
LGTM, I can see from here it looks like this infrastructure matches the meta state https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data-infrastructure/actions/runs/24787218616/job/72534499186

terraform/meta/.terraform.lock.hcl
natdeanlewissoftwire: tf just hadn't been re-`init`ed since this dependency was added, confirmed even with `terraform init -reconfigure` this should still be present

